PR #304, opened 2026-04-16
feat(sync): forced and project-scoped task sync import

anti-duhring:
## Summary
Add forced and project-scoped task sync import.

## What changed
- Added `force` to `task_sync_import`
- Made `task_sync_import(project=...)` import only the requested project
- Added `import_project(project_id, force)` through the sync bridge/manager
- Updated task sync import to bypass the `last_imported_sha` short-circuit when `force=true`
- Kept background/normal imports unchanged (`force=false`)

## Why
This fixes recovery scenarios where `djinn/tasks` is the source of truth but local task state is missing or stale. Previously, import could silently no-op if the remote branch SHA matched the stored `last_imported_sha`, and the `project` parameter was accepted but not actually used to scope the import.

## Result
You can now force a rebuild-style import for a single project, while preserving the existing SHA-skip optimization for normal sync flows.

### DIFF
--- a/server/crates/djinn-agent/src/context.rs
+++ b/server/crates/djinn-agent/src/context.rs
@@ -213,8 +213,12 @@ impl bridge::SyncOps for AgentSyncOps {
         vec![]
     }
 
-    async fn import_all(&self) -> Vec<bridge::SyncResult> {
+    async fn import_all(&self, _: bool) -> Vec<bridge::SyncResult> {
         vec![]
+    }
+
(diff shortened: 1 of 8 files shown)


fernandobandeira:
Closing as stale — thank you for the contribution. This extends the `djinn/tasks` sync-branch subsystem, which was intentionally removed from the server in `5584c0ca` (2026-04-20). The files this PR edits no longer exist on main, so it's unmergeable and has no integration target. No reflection on the work itself — the underlying feature was deprecated.

PR #306, opened 2026-05-26
feat(k8s,chart): configurable nodeSelector + tolerations for task-run + warm Pods

thales-altpay:
## Motivation

The djinn-k8s controller builds Job PodSpecs for task-run and warm workloads but doesn't expose any scheduling hints. On multi-NodePool clusters, those Pods fall through to whatever default pool can take them — even when the operator has provisioned a dedicated NodePool (e.g. a Karpenter pool labelled/tainted `workload-type=djinn`) specifically for djinn builds.

Today the only way for a Pod to land on such a pool is for it to:
- carry a matching `nodeSelector` so the scheduler picks it, *and*
- tolerate the pool's `NoSchedule` taint so it isn't rejected at admission.

Neither is configurable without a code patch.

## Changes

Two new fields on `KubernetesConfig`, **shared between task-run and warm Pods**:

```rust
pub node_selector: BTreeMap<String, String>,
pub tolerations: Vec<k8s_openapi::api::core::v1::Toleration>,
```

Warm Pods pre-warm the canonical-graph cache that task-runs later adopt, so they have to land on the same pool — otherwise the warmup is wasted. A single set of scheduling hints covers both.

Surfaced two ways, matching the existing chart pattern:

- **Env vars on the controller** (JSON-encoded): `DJINN_K8S_NODE_SELECTOR`, `DJINN_K8S_TOLERATIONS`. Parsed by `KubernetesConfig::from_env()`; malformed JSON falls back to the default with a `tracing::warn` (same pattern as the existing `DJINN_K8S_TTL_SECONDS` parser).
- **Helm chart values**: `resources.taskrun.nodeSelector` and `resources.taskrun.tolerations`. Both default to empty (`{}` / `[]`) so the rendered manifest is byte-identical to the pre-feature shape for any existing install.

Example values:

\`\`\`yaml
resources:
  taskrun:
    nodeSelector:
      workload-type: djinn
    tolerations:
      - key: workload-type
        operator: Equal
        value: djinn
        effect: NoSchedule
\`\`\`

## Files

| File | Change |
|---|---|
| `server/crates/djinn-k8s/src/config.rs` | Add fields, defaults, env parsing, doc-table entry, happy-path unit test |
| `server/crates/djinn-k8s/src/job.rs` | Import `Toleration`; set `node_selector` + `tolerations` on the task-run PodSpec (both `None` when unset → same shape as before) |
| `server/crates/djinn-k8s/src/warm_job.rs` | Same wiring on the warm-Job PodSpec |
| `deploy/helm/djinn/values.yaml` | New `nodeSelector` / `tolerations` keys under `resources.taskrun`, with a comment block explaining the use case |
| `deploy/helm/djinn/templates/deployment-server.yaml` | Render the two new env vars on the controller, guarded by `{{- with }}` so empty defaults emit nothing |

## Verification

- `helm lint` passes.
- `helm template` with the keys set renders the env vars with correct JSON.
- `helm template` with defaults emits no `DJINN_K8S_NODE_SELECTOR` / `DJINN_K8S_TOLERATIONS` lines — backward-compatible.
- New unit test (`from_env_parses_pod_scheduling_json`) covers happy-path JSON deserialization for both env vars, following the existing `ENV_LOCK`/save-restore pattern.
- `cargo check` / `cargo test` not run locally in this PR — relying on CI for compile + test verification.

## Scope deliberately punted

- **`affinity`**: not included. nodeSelector + tolerations covers the common NodePool-pinning case; affinity has a much larger surface (preferred vs required, multiple terms, complex matchers) and deserves its own design discussion.
- **Separate scheduling for warm vs task-run**: rejected because warm Pods exist to pre-warm caches the task-run adopts. Diverging their pools defeats the purpose. If a future use case needs them split, this can grow into `resources.warm.nodeSelector` later without a breaking API change.

### DIFF
--- a/deploy/helm/djinn/templates/deployment-server.yaml
+++ b/deploy/helm/djinn/templates/deployment-server.yaml
@@ -163,6 +163,20 @@ spec:
                         (include "djinn.fullname" .)
                         (include "djinn.namespace" .)
                         (.Values.service.rpcPort | int) | quote }}
+            # Pod scheduling for task-run + warm Jobs. JSON-encoded so the
+            # controller can deserialize directly into k8s_openapi types
+            # without bringing a YAML parser into the dispatch path. The
+            # `with` blocks omit the env var entirely when the values are
+            # empty defaults, leaving the controller on its built-in fallback
+            # (no scheduling hints — matches pre-feature behavior).
+            {{- with .Values.resources.taskrun.nodeSelector }}
+            - name: DJINN_K8S_NODE_SELECTOR
+              value: {{ toJson . | quote }}
+            {{- end }}
+            {{- with .Values.resources.taskrun.tolerations }}
+            - name: DJINN_K8S_TOLERATIONS
+              value: {{ toJson . | quote }}
+            {{- end }}
             - name: DJINN_VAULT_KEY_PATH
               value: /var/run/secrets/djinn/vault.key
             - name: GITHUB_APP_PRIVATE_KEY_PATH
--- a/deploy/helm/djinn/values.yaml
+++ b/deploy/helm/djinn/values.yaml
@@ -85,6 +85,26 @@ resources:
     limits:
       cpu: "2"
       memory: "4Gi"
+    # Pod scheduling for both task-run and warm Jobs. Warm Pods pre-warm the
+    # graph cache that task-runs later adopt, so they MUST land on the same
+    # NodePool — that's why one set of hints covers both.
+    #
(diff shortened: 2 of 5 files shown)


fernandobandeira (approved):

thales-altpay:
I have read the CLA Document and I hereby sign the CLA.

fernandobandeira:
recheck

fernandobandeira:
recheck

PR #618, opened 2026-06-15
fix(board): populate merge_commit_sha so merged tasks show again

fernandobandeira:
## Problem

The board's **Merged** column was gated on `task.merge_commit_sha != null` (#572), and the field was threaded through the list/SSE payload contracts (#574) — but **nothing ever populated it** on the actual PR-merge path. `set_merge_commit_sha` was only ever called from a proposal tool and a test. So every closed task had a `NULL` SHA and the board hid **all** merged tasks. The same field gates `record_merge_event`, so throughput metrics were silently broken too.

## Fix

**Backend — persist the landed SHA on every merge-close path**
- Add `merge_commit_sha` to the `PullRequest` GitHub API type (it was being dropped on deserialization).
- New `apply_pr_merge(task_id, sha)` helper persists the SHA **before** the `PrMerge` transition, so the `task_updated` event the transition emits already carries it (board placement + throughput in one consistent state).
- Wire the three production merge-close sites through it: both `pr.merged == true` detections (`pr_watcher` / `pr_review_watcher`) use `pr.merge_commit_sha`; the squash-merge success uses the `sha` from the `PUT /merge` response (previously discarded as `Ok(_)`). Merge-queue success closes via the same `pr.merged` path.

**Frontend — backward-compat for rows closed before the SHA was persisted**
- A closed task counts as Merged when it has a SHA, **or** (legacy) it opened a PR (`pr_url`) and closed as `completed`. This excludes force-closed-unmerged (`force_closed`) and the review/decomposition/planning/spike tasks that never open a PR.

## Tests
- Provider: a merged PR deserializes `merge_commit_sha`.
- Board: legacy PR-completed task shows as Merged; force-closed task stays hidden.

All affected crates compile; provider + board tests green (`tsc` clean).

### DIFF
--- a/server/crates/djinn-agent/src/actors/coordinator/pr_poller/pr_review_handlers.rs
+++ b/server/crates/djinn-agent/src/actors/coordinator/pr_poller/pr_review_handlers.rs
@@ -327,6 +327,38 @@ impl CoordinatorActor {
         }
     }
 
+    /// Close a task whose PR has merged, recording the landed merge-commit SHA.
+    ///
+    /// The SHA is persisted *before* the `PrMerge` transition so the
+    /// `task_updated` event the transition emits already carries it — that is
+    /// what lets the board place the card in the Merged column and the
+    /// coordinator record a throughput event (both gate on
+    /// `merge_commit_sha IS NOT NULL`). An empty/absent SHA degrades to a plain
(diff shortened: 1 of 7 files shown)


fernandobandeira:
I have read the CLA Document and I hereby sign the CLA

PR #772, opened 2026-06-18
perf(ci): build test binaries once via nextest archive, run shards from artifact

fernandobandeira:
## Build-once, run-many

The `server-test` job ran as a 4-way matrix where **each** shard recompiled the full test suite (incrementally off rust-cache) before running its `--partition count:N/4` slice — up to 4x duplicate compilation.

This switches to the **`cargo nextest archive`** build-once-run-many pattern:

- **New `server-test-build` job** compiles the whole suite **once** (`cargo nextest archive --workspace --all-targets --all-features --profile ci`) and uploads the archive as an artifact.
  - The build needs **no Postgres**: `sqlx::query!` macros resolve against the committed `server/.sqlx/` offline cache with `SQLX_OFFLINE=true` (freshness gated separately by the `Server sqlx Cache` job), exactly as `server-clippy`/warm-cache already build. Only the test **runtime** needs a DB.
- **`server-test` matrix** now downloads that archive and runs the prebuilt binaries (`cargo nextest run --archive-file ... --workspace-remap . --partition count:N/4 --profile ci`). No rust-cache restore, no compile. All runtime setup (postgres service, migrations, test template, vault key, git identity) is preserved.
- **Quality Gate** gains `server-test-build` in its `needs`/results array so a build failure also gates the merge.

Net effect: the test suite compiles **once** instead of up to four times across the shards.

### DIFF
--- a/.github/workflows/quality-gate.yml
+++ b/.github/workflows/quality-gate.yml
@@ -272,9 +272,57 @@ jobs:
   # ║  test build when clippy would have failed anyway.                ║
   # ╚══════════════════════════════════════════════════════════════════╝
 
+  # ── Build test binaries ONCE ─────────────────────────────────────────
+  # Compile the full test suite a single time into a nextest archive, then
+  # hand it to the sharded `server-test` matrix below. This replaces the
+  # old pattern where each of the 4 shards recompiled the test binaries off
+  # the restored rust-cache (up to 4× duplicate compilation).
+  #
+  # The build needs NO database: `cargo nextest archive` compiles the
+  # `sqlx::query!` macros, but with SQLX_OFFLINE=true + the committed
+  # `server/.sqlx/` cache (kept fresh by the `Server sqlx Cache` job) they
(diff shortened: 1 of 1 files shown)


fernandobandeira:
Dropping in favor of the per-shard-compile sharding already merged in #771 (which works).

The nextest-archive build-once approach is blocked by 3 worker integration tests (`djinn-agent-worker/tests/{rpc_roundtrip,cancel_path,in_pod_drive}.rs`) that resolve the worker binary via the **compile-time** `env!("CARGO_BIN_EXE_djinn-agent-worker")`, which bakes in the build-job path and can't see nextest's runtime `NEXTEST_BIN_EXE_*` on the shard runners → `spawn worker: NotFound`.

Even with that fixed (a 3-line runtime resolver), the archive is ~1.2 GB → uploaded once + downloaded by 4 shards (~6 GB transfer) + extract overhead, which likely offsets the compile-once savings unless the test binaries are also stripped. Not worth the complexity right now; #771's sharding already gives the parallelism win.

PR #1882, opened 2026-07-11
Enforce deterministic merge-queue gates before PR open (incl. Migrations Guard)

fernandobandeira:
## Problem

A production audit today found **45% of task reopen events are merge-queue rejections with repeated identical failure fingerprints** — deterministic checks that only fail on the `merge_group` base ref, so a PR looks green, the reviewer approves, and then the queue rejects. Observed repeat offenders: **Server sqlx Cache, Server Size Guard, Migrations Guard, Raw-SQL Boundary, Quality Gate**. Each rejection routes through lead-intervention → reopen → redispatch (11–58 min waits) and burns a whole new worker session, often hitting the same gate 3–5 times in a row.

## Root cause

The pre-approval CI-grade verification gate (`preapproval_gate.rs`, proposal `uv3p`) already mirrors most PR-level `server` checks and runs at `supervisor_pr_open` *before the task branch is pushed*. But it had two gaps:

1. **`SERVER_CHECK_SET` omitted the Migrations Guard** — a named repeat offender. A migration-immutability violation sailed through the gate entirely and only detonated in the merge queue.
2. **It ran purely in shadow mode.** `inline_exec` is OFF by default, so on a cache miss the gate returned `DeferredNoVerdict` and proceeded; `run_check_set` was the only thing that ever populated the cache. Net effect: the deterministic offenders were never actually blocked pre-PR.

## Change

- **Add the Migrations Guard** to `SERVER_CHECK_SET`, reusing the exact CI script (`scripts/check-migrations-immutable.sh`) — no reimplemented logic to drift.
- **Tag each check `requires_build`** and split execution: the **no-compile deterministic guards** (Size Guard, Migrations Guard, Raw-SQL Boundary, capability boundaries) now run **inline even in shadow mode**, so a red guard blocks PR-open **strike-free** (feedback on the `verification` channel, task returns to a worker round — no reopen/intervention counter touched). The compile/DB-heavy checks (clippy, sqlx cache, test build) still defer to the verification pod unless `inline_exec` is enabled.
- **False-block guard:** the no-compile guards diff against `origin/main`; if the worktree has no resolvable base the gate falls back to the historical deferral instead of misreading "no base" as a violation.
- **`scripts/premerge-gates.sh`** — a single locally-runnable reproduction of the deterministic gates (Size Guard, Migrations Guard, Raw-SQL Boundary, capability boundaries, architectural boundaries; `--with-sqlx` adds the sqlx-cache check). It reuses the same CI scripts byte-for-byte, so a worker/human can catch these before requesting review. A drift-guard test ties it to the Rust check set.

Everything delegates to the same `scripts/*` / `make` targets CI runs — the Rust `SERVER_CHECK_SET` and the shell script stay in lockstep via `check_set_scripts_exist`, `no_build_subset_is_the_deterministic_guards`, and `premerge_gates_script_covers_no_build_subset` tests.

## Validation

- `scripts/premerge-gates.sh` passes on clean `main`; with a crafted migration edit + a raw-`sqlx::query!` outside `djinn-db`, it reports `FAIL migrations_guard` / `FAIL raw_sql_boundary` and exits 1.
- `cargo test -p djinn-coordinator preapproval_gate` — 16/16 pass, including new shadow-mode tests: green no-compile subset defers (no partial-green cache row), red no-compile subset blocks strike-free, and an unresolvable base defers without running any gate.
- `cargo clippy -p djinn-coordinator --all-targets` clean; `cargo fmt --check` clean.


fernandobandeira:
Closing without merge: wrong direction. This doubles down on an in-house, djinn-repo-specific CI mirror (hardcoded scripts in the coordinator) that cannot generalize to other projects. Replacement: open task PRs as drafts, let the project's own GitHub Actions run on the PR head, feed failing check results back to the worker, and mark ready-for-review only when green. The preapproval_gate machinery will be removed entirely.

PR #2572, opened 2026-07-25
fix(8hrv): classify nextest TRY-n FAIL lines in shard failure summaries

fernandobandeira:
## The defect

The Server Test shard's `Surface shard failure at run level` step grepped the tee'd nextest log for `^[[:space:]]*FAIL \[`. When that matched nothing it emitted a message that **asserted a cause**:

> Server Test shard-1 failed — shard concluded failure without nextest FAIL lines (build, migration, or setup step); see this job's log.

That annotation is what the board surfaces as failure evidence (wnqw / #2561 added annotation capture), so nineteen remediation sessions on task `rwpk` (#2565) were told — authoritatively and wrongly — that the failure was in a build/migration/setup step. The real failure was an ordinary assertion at `shard1.log:4258`:

```
djinn-provider::stream_event_consumer_audit
  checked_in_classification_covers_every_production_stream_event_match
assertion `left == right` failed
```

## Why the pattern could not match

Two independent reasons, both confirmed against the captured shard log of run `30149467240` / job `89657358613` and reproduced locally with `cargo-nextest 0.9.133`:

1. **Retries.** `pull-request`, `merge-group` and `full-validation` all inherit `[profile.ci]`'s `retries = { count = 2, ... }` (`server/.config/nextest.toml`), so a genuinely failing test never prints a bare `FAIL` line. Every attempt, the final one, and its repeat in the Summary block are prefixed:

   ```
   TRY 3 FAIL [   0.483s] (2477/2842) djinn-provider::stream_event_consumer_audit checked_in_…
   ```

2. **Color.** `quality-gate.yml` sets `CARGO_TERM_COLOR: always` workflow-wide, so nextest colorizes into the pipe and the escape wraps the *indentation together with the status token*:

   ```
   \e[31;1m        FAIL\e[0m [   0.002s] (1/3) crate test_name
   ```

   The old pattern's leading `^[[:space:]]*` anchor and its space before `[` are therefore both wrong even for an un-retried failure — the inline grep could never have matched on CI at all.

A third latent bug: nextest prints a progress counter after the timing bracket (`(2477/2842)`, or `(───)` before the total settles). The old `sed` stripped only the timing bracket, so had it ever matched it would have emitted `(2477/2842) <binary-id> <test-name>` as the test id.

## Pattern before → after

Before (inline in the workflow, untestable):

```sh
grep -E '^[[:space:]]*FAIL \[' "$log" \
  | sed -E 's/^[[:space:]]*FAIL \[[^]]*\][[:space:]]+//' \
  | sort -u
```

After — `scripts/ci-shard-failure-summary.mjs`, escapes stripped first, then:

```
^{<log decoration> }?[indent]{TRY <n> }?<STATUS> [<duration>] {(<progress>) }?<binary-id> <test-name>
```

with `STATUS ∈ {FAIL, LEAK-FAIL, TIMEOUT, ABORT, SIG*}` counted as a failure and `{PASS, LEAK}` used only to **retract** a candidate (a flaky test prints `TRY 1 FAIL` then `TRY 2 PASS`; blaming the shard on it is the same misattribution pointed the other way). First-report order is kept and duplicates removed. `LEAK-FAIL`, `SIGABRT`, `SIGSEGV` were verified to exist as status tokens in the pinned nextest build.

The parsing lives in a checked-in script rather than inline shell precisely because the inline version could not be tested and was wrong in two ways at once.

## Message rewording

The generic branch no longer names a layer it cannot see. It reports what it observed, and quotes evidence:

```
::error title=Server Test shard-2 failed::nextest ran and its log (…/nextest-run.log) holds no
failure status line (FAIL / TRY n FAIL / TIMEOUT / SIG*). First error line in that log: "linking
with `cc` failed: exit status: 1". This does not identify which step failed — read the last red
step in this job's log.
```

The job-page summary adds a bounded, fenced tail of the last 15 non-empty log lines. Quoted text is clamped at 400 chars per line. An absent log is reported as an absent log, not as a build failure.

## Fixtures — both directions

`scripts/ci-shard-failure-summary.test.mjs` runs in preflight (`node --test`) with fixtures under `scripts/fixtures/shard-failure/`:

| direction | fixture | asserts |
| --- | --- | --- |
| TRY-n only ⇒ **test failure** | `nextest-try-n-fail.log` | yields exactly `djinn-provider::stream_event_consumer_audit checked_in_…`; annotation and summary contain no `build`/`migration`/`setup` text at all, and the generic diagnostic is not selected |
| same, operator's copy | `nextest-try-n-fail-gh-log-capture.log` | identical classification from `gh run view --log` bytes (job/step/timestamp decoration, `^[` transliterated escapes) |
| no FAIL of any kind ⇒ **generic** | `nextest-no-failure-lines.log` | empty failing set, generic wording selected, first `error:` line quoted, log tail present, and the old cause-asserting phrasing absent |
| ordinary FAIL + dedup | `nextest-plain-fail.log` | two ids in report order, each present twice in the log (live + Summary), deduplicated |
| flaky recovered | `nextest-flaky-recovered.log` | `TRY 1 FAIL` + `TRY 2 PASS` ⇒ empty failing set |
| absent log | – | reported as an absent log |

The first two fixtures are whole lines lifted **verbatim** from the captured shard log of the misleading run — provenance was verified line-by-line against the capture, and the test file documents which fixtures are captured and which are constructed. The remaining three are constructed, but every status line in them is a shape reproduced locally from `cargo-nextest 0.9.133` with `CARGO_TERM_COLOR=always`.

The test also pins the workflow contract: the shard step must call the script, preflight must run the test, and neither the old grep nor the phrase `shard concluded failure without nextest FAIL lines` may come back.

## Verification

- `node --test scripts/ci-shard-failure-summary.test.mjs` — 14/14 pass.
- End-to-end against the **full 6497-line captured shard log**: emits exactly the one true failing test, no false positives.
- Against three logs produced locally by `cargo-nextest 0.9.133` (`--profile ci`, `--profile pull-request`, and `--retries 0` with color forced): correct in all three, including flaky retraction.
- `actionlint` clean on `quality-gate.yml`; YAML parses.
- No Rust touched, so the workspace gates are unaffected.

Closes `8hrv`.

### DIFF
--- a/.github/workflows/quality-gate.yml
+++ b/.github/workflows/quality-gate.yml
@@ -142,6 +142,8 @@ jobs:
         run: node --test scripts/ci-qa-smoke-workflow.test.mjs
       - name: Verify release image validation contract
         run: node --test scripts/ci-release-image-validation.test.mjs
+      - name: Verify shard failure summary contract
+        run: node --test scripts/ci-shard-failure-summary.test.mjs
       - name: Verify retirement manifest generator contract
         run: node --test scripts/test-djinn-retirement-manifest.mjs
       - name: Run retirement manifest strict reconciliation guard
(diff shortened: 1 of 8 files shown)


fernandobandeira:
Superseded by #2571, which djinn produced first and which is already green. #2571 additionally distinguishes a flaky-recovered test (TRY 1 FAIL then TRY 2 PASS) from a genuine failure, and handles the ANSI-escaped TRY-n form — both beyond what this branch covered. Closing to avoid two implementations of the same shard-summary parser.

PR #2574, opened 2026-07-25
fix(9jrg): make $HOME writable by the uid the pod runs as, and take the durable stash off it

fernandobandeira:
## P0

Every worker and planner session on deployed v0.7.3 died with
`reply loop error: create durable blobs: Permission denied (os error 13)`
(13 worker + 7 planner in a 40-minute window). Confirmed in a live pod:
`uid=1000 gid=1000`, `HOME=/home/djinn`, `drwxrwxr-x djinn djinn`,
`mkdir /home/djinn/.cache -> Permission denied`.

`qut0` moved task-run Pods to uid/gid 1000 while the image still owned
`/home/djinn` as `10001:10001` mode `0775`. The group-write bit was there but
pointed at a group the Pod does not hold, so the Pod fell through to "other"
(`r-x`) and could not create a single entry under its own `$HOME`. The durable
output stash resolves `$XDG_CACHE_HOME/djinn/output_stash` and falls back to
`$HOME/.cache/djinn/output_stash`; `XDG_CACHE_HOME` was unset in the rendered
Job.

Knock-on: a worker dies before submitting for review, so the task returns to
`open` and no reviewer is ever spawned. Reviewer dispatch was never broken, just
never reached. After six consecutive failures the coordinator escalates to a
1800s cooldown, which makes planning look dormant rather than failing.

## Layer chosen: both, because they answer different questions

**Job render (`djinn-k8s`)** — `XDG_CACHE_HOME=/cache/xdg/<project_id>`.

The env route is not just a permission workaround here. Even with the image's
ownership fixed, `$HOME/.cache` is a **container layer that dies with the Pod**,
so a stash that is read back after a restart and expires on a retention window
(`DJINN_OUTPUT_STASH_GC_RETENTION_DAYS`) had no business living there. `/cache`
is:

- persistent, so cross-run persistence is *gained*, not silently dropped — and
  the dir is deliberately NOT keyed by task run;
- group-owned by the artifact GID with setgid `2775`, so `create_dir_all` under
  the worker's `0002` umask inherits a conforming subtree, and writability is
  **verified** at startup by the volume contract rather than assumed;
- inside the agent's Landlock allowlist automatically — the sandbox derives its
  djinn cache dir from this same env var;
- namespaced per project like `SCCACHE_DIR`/`CARGO_TARGET_DIR`: the PVC is shared
  cluster-wide and stashed blobs are verbatim tool output.

Set in the shared cache-env helper, so warm Pods get it too — the SCIP indexer
cache (`djinn-graph/scip_indexer/cache.rs`) resolves the same env pair and was
silently broken by the same migration.

**Image** — `/home/djinn` keeps uid `10001` as its **owner** and becomes
group-owned by the artifact GID `1000` with setgid `2775`.

This is the reconciliation the dual-uid problem needs. Three identities write
that directory: uid `10001` (the server-side path), uid/gid `1000` (the worker
since `qut0`), and uid `1001`/group `1000` (the launcher-spawned child that runs
the agent's shell commands). Only a shared **group** can hold all three;
`chown`ing to 1000 would have broken the first. This is the fix for the wider
class the task calls out — `fnm`'s multishell state, `gopls`, `npm` and
`git config --global` all sit on the same wall. Applied to both
`base-debian.sh` (project images) and `djinn-agent-runtime-base.Dockerfile`,
with the hash salt bumped to `env-config/v11` so every catalog image rebuilds.

## Readiness check: added

`volume_contract::check_home_writable` asserts the running identity can create
entries in `$HOME`, failing readiness with the path, the observed
ownership/mode, and the running identity (uid, gid, supplementary groups).

`$HOME` is deliberately **not** modelled as a `VolumeRoot`: it is an image path,
the artifact-gid/setgid rules do not apply to it, and it is correct for it to
stay owned by uid 10001. Only its *effective* writability is contractual — so the
check asks the kernel (`access(W_OK|X_OK)`), which accounts for the
owner/group/other selection, supplementary groups, ACLs and a read-only mount.
It creates nothing: a probe file would have to be written into the directory
whose writability is in doubt.

It fails **closed**, and that is safe by construction: Pods only start on a
`ready` catalog image, and the salt bump marks every catalog image `building`
until it rebuilds with the fixed `$HOME`, so a pre-fix image cannot be dispatched
to. `DJINN_VOLUME_CONTRACT_MODE=audit` remains the existing loud break-glass.

## Tests never run as the owner

A test that owns the directory it probes passes for the wrong reason — the owner
class carries `rwx` in `0775`. That is the gap that let this, `ej9c` and `4hfr`
ship, so `tests/home_contract.rs` has two arms and one of them always runs:

- **Unprivileged** (any CI runner): probes a real directory the filesystem
  already has owned by another uid with no group/other write, asserts the typed
  rejection records `observed_uid != process_uid`, and asserts the genuine
  `create_dir_all(".cache/djinn/output_stash/blobs")` fails with the production
  `PermissionDenied` — the check's verdict has to match the kernel's. Then
  asserts `enforce_with` fails closed on it.
- **Privileged** (root in a container): fabricates both real shapes —
  `10001:10001 0775` (the outage) and `10001:1000 2775` (the fix) — and judges
  each from a child `fork`ed and dropped to uid/gid 1000 with supplementary group
  1000, running both the check and a real stash `create_dir_all`. This proves the
  bug *and* the image fix in one test.

If neither arm is possible the test fails rather than passing vacuously.
Verified by mutation: reverting the fixed shape's group to `10001` fails it
(`must be accepted by 10001:1000 2775 ... (exit 2)`), and verified the privileged
arm actually executes (`docker run -u 0 archlinux:base <test-bin>`).

## Follow-up, recorded not hidden

The coordinator's stash GC runs in the server process against the server's own
root and does not sweep the pod-side root. Neither did anything before — the
pod-side stash lived on a container layer that died with the Pod, so it was never
collected *or* persisted. The server mounts this same claim at
`/var/lib/djinn/cache`, so extending that sweep to the per-project roots needs no
new volume plumbing. Noted at the render site and in the runbook.

## Gates

- `SQLX_OFFLINE=1 cargo check --workspace --all-targets` — clean
- `cargo fmt --all --check` — clean
- `cargo clippy -D warnings` on djinn-agent-worker / djinn-k8s / djinn-image-builder — clean
- `cargo nextest`: djinn-k8s + djinn-image-builder 272 passed, djinn-agent-worker 363 passed, djinn-agent 1428 passed, djinn-coordinator 1934 passed
- file-size guard (1500 lines / 51,200 bytes), raw-SQL boundary, k8s boundary, shared-cache rollout (112 checks) — all pass
- no `Cargo.toml` changed, so no hakari regeneration needed

## Deploy note

The server change lands the fix on the next dispatched task-run with no image
rebuild. The `$HOME` readiness check and the image fix arrive together: catalog
images go `building` on the salt bump and Pods will not dispatch until they are
`ready` again.

### DIFF
--- a/docs/VOLUME_OWNERSHIP_CONTRACT_RUNBOOK.md
+++ b/docs/VOLUME_OWNERSHIP_CONTRACT_RUNBOOK.md
@@ -91,6 +91,55 @@ the startup check **never repairs** — a recursive `chown` over a 300G cache is
 multi-minute stall on pod start, which is the whole reason
 `fsGroupChangePolicy` is `OnRootMismatch`.
 
+## `$HOME` is part of the startup check too (9jrg)
+
+The volume contract covers the *mounted* surfaces. It did not cover the one
+writable surface that ships inside the image: `$HOME` (`/home/djinn`).
+
+`qut0` moved the Pod to uid/gid `1000` while the image still owned `/home/djinn`
+as `10001:10001` mode `0775`. The group-write bit was there but pointed at a
+group the Pod does not hold, so the Pod fell through to *other* (`r-x`) and could
+not create a single entry under its own home. Nothing checked it. The first
+consumer to notice was the durable output stash — `$HOME/.cache/djinn/output_stash`
+when `XDG_CACHE_HOME` is unset — and it surfaced hours later, inside the reply
+loop, as `create durable blobs: Permission denied (os error 13)` on **every**
+worker and planner session. Because a worker dies before submitting for review,
+the task returns to `open` and no reviewer is ever spawned; after six consecutive
+failures the coordinator escalates to a 1800s cooldown, so planning looks dormant
+rather than failing.
+
+Two changes close it:
+
+| Layer | Mechanism |
+|-------|-----------|
+| Image | `/home/djinn` keeps uid `10001` as its **owner** and is group-owned by the artifact GID `1000` with `2775`. Three identities write it — uid `10001` (server-side path), uid/gid `1000` (worker), uid `1001`/group `1000` (launcher-spawned child) — and only a shared group can hold all three. Changing the owner instead would break the first. Applied in `djinn-image-builder/scripts/base-debian.sh` (project images, `env-config/v11` forces the rebuild) and `server/docker/djinn-agent-runtime-base.Dockerfile`. |
+| Job render | `XDG_CACHE_HOME=/cache/xdg/<project_id>` — the stash and the SCIP indexer cache resolve `$XDG_CACHE_HOME` first, so they no longer depend on the image home at all, and they land on a *persistent* PVC instead of a container layer that dies with the Pod. |
+| Runtime | `volume_contract::check_home_writable` asks the kernel (`access(W_OK|X_OK)`) whether the running identity can create entries in `$HOME`, and fails readiness with the path, observed ownership/mode and running identity. |
+
+The same rule applies to anything the image pre-creates *under* `$HOME` for the
+runtime identity to write. `install-node.sh` scaffolds `$HOME/.local/state` so
+`fnm` can drop its per-shell multishell symlink there; `mkdir` under the build
+umask leaves it `0755`, which stopped working for the same reason the moment the
+runtime identity stopped being the owner. It is now `2775` too. When adding a
+build-time directory the Pod must write, group-own it to `1000` and give it
+`2775` — owner-only modes are only safe for paths nothing writes at runtime.
+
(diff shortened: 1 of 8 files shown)


fernandobandeira:
Amended to also fix `$HOME/.local/state` in `install-node.sh` (`chmod 2775`). That block exists so `fnm` can write its per-shell multishell symlink, but `mkdir` under the build umask left it `0755` — which stopped working the moment the runtime identity stopped being the owner, exactly like `/home/djinn` itself. Same rebuild, same class; documented in the runbook as the rule for any build-time directory the Pod must write.

PR #2600, opened 2026-07-25
fix(spec): Correct the Light role-class falsehood and make compile pressure measurable

fernandobandeira:
## What was wrong

`RoleResourceClass::for_role` documented `Light` as *"Orchestration-only roles that never run the project's compile/test toolchain"*.

That claim is false, and was already known to be false when it landed. Measured over production transcripts on 2026-07-25:

- **4 of 73** light task-run sessions (**5.5%**) ran a real compile.
- **3 of 37** reviewer sessions (**8.1%**). One ran `cargo check -p djinn-agent-worker`, `cargo clippy --all-targets` *and* `cargo test --no-run` in a single session, with `Compiling djinn-agent-worker` in the output, citing `task-reviewer.md:22` as authorization.
- Light roles are **56%** of all task-runs; reviewers alone are 42%.
- Commit `1719ef8c3` (2026-06-20) had already recorded a reviewer burning ~12 minutes on a cold `cargo check` *"despite task-reviewer.md already instructing it not to"*. The `RoleResourceClass` comment asserting the opposite landed **34 days later**.

## What is deliberately unchanged

**No behavioral change to who is gated at dispatch.** The architecture was already correct, and this PR does not reclassify anyone.

Compile pressure is governed in two layers:

1. **Dispatch admission** — coarse, role-derived, `RoleResourceClass`. Pre-charges a scarce build slot only to work *certain* to compile. Light roles are exempt, and that is right: paying a slot 100% of the time for a ~5% event, on an oversubscribed pool, is the wrong trade.
2. **The invocation lease** — fine, measured, **role-agnostic**. `LeaseInvocationRunner` queues purely on measured `cpu.stat` (`process.rs`, threshold in `context.rs`), and `BuildLeaseService::queue` is keyed only by `{task_id, task_run_id, invocation_id}`. A reviewer's compile already takes a lease on exactly the same terms as a worker's.

So the falsehood was only ever in the *justification text* and in a contradictory second classifier. Both are fixed here.

## Changes

**1. `djinn-runtime/src/spec.rs` — the doc comment.** Rewritten around the measured numbers and the two-layer model. `Light` now means "unlikely to compile, so not pre-charged a slot", not "cannot compile". Cites `1719ef8c3`.

**2. `consumes_build_slot()` -> `gated_at_dispatch()`.** The old name implied a capability; the new one names the one thing it decides. Documented explicitly: **there is no `may_take_invocation_lease()` companion, on purpose.** It would return `true` for every input — a constant dressed as a classification, and an open invitation for a future author to make it role-dependent. "Everyone may lease" is encoded as a documented absence plus two tests.

Sole caller in `build_admission.rs` is a **one-line** rename (another agent is working in that file on admission-journal recovery; rebased onto `main` immediately before pushing).

**3. `task-reviewer.md` — a self-contradiction.** Line 23 forbade build/lint/test suites while line 22 offered `cargo modules dependencies` as an inspection example — and that command *compiles the crate*. Replaced with examples that read what is already on disk (`grep`/`rg`, `git log`, `git diff --stat`, `cargo tree`, `cargo metadata`, reading a generated file), plus an explicit note that `cargo modules ...`/`cargo expand`/`cargo doc` are builds. Added the exception plainly: clippy/test are permitted **only** to verify an AC that names them, scoped with `-p`.

**4. `workspace.rs` — the duplicate classifier is gone.** It was:

```rust
if matches!(session_role, Some("worker") | Some("reviewer"))
```

with a comment claiming *"Other roles (planner/architect) don't run cargo"*. It disagreed with `spec.rs` in **both** directions: it classed Reviewer *with* Worker, and it excluded Architect, which `spec.rs` classes `BuildCapable`. `RoleResourceClass::for_role_name` is now the only role classifier in that file.

**Behavioral widening, called out explicitly:** the `cargo check`/`cargo build` -> clippy steer now applies to **every** role — architect and verifier as intended, and also planner/lead/refinement and an absent role.

I did **not** key it on `class == BuildCapable`. That would have *removed* the steer from Reviewer, the one Light role actually measured to compile, reintroducing the exact cold-`cargo check` pathology from `1719ef8c3`. The steer is a property of the **pod's cache shape** (clippy-warmed), not of the role — which is this PR's own thesis applied consistently. Covered by `cargo_check_steer_applies_to_every_role`, which asserts the steer for all eleven role spellings including `None`.

**5. Observability — the part that actually blocks arming.** Added a bounded two-value `class` label (`RoleResourceClass::as_str()`) to `cargo_invocation::record_seconds` and `build_admission::record_shadow_invocation`.

### Telemetry transport: structured `tracing::info!`

`djinn-agent-worker` calls `djinn_telemetry::init()` (`main.rs:393`) but **serves no HTTP `/metrics` endpoint at all**. Every caller of both metrics runs *inside* the task-run pod, so both were recorded in-pod and never scraped — which makes "what fraction of invocations would escalate?" unanswerable and a shadow rollout meaningless.

**Chosen: emit a structured `tracing::info!` alongside each metric**, collected by the pod-log pipeline that already exists. It is strictly the smaller of the two options — standing up an in-pod HTTP server, or widening the host RPC and its wire types, both cost more and add a surface that can fail closed. It also follows the precedent already in the worker (`cargo_metrics.rs` emits log + metric side by side for every cache counter). The histogram/counter are kept: the same code path is exercised by host-side tests via `djinn_telemetry::render()`.

The label reaches the two recorders differently, for a deliberate reason:

- `cargo_invocation::record_seconds` takes `class` **explicitly**, threaded from `session_role` at `workspace.rs` through `finish_shell`.
- `record_shadow_invocation` reads a **process-published** class (`role_class::observe`/`current`). A task-run pod hosts exactly one role, so the class is a process constant — and threading a role into `LeaseInvocationRunner` just to label a counter would put a role-shaped field exactly where a future author would reach to make leasing role-dependent. The runner stays blind. Unobserved reads as `build-capable`, so an unlabelled process never under-reports compile pressure.

**6. `"grooming"` — removed.** `for_role_name` listed it as a light role name, but `TaskRunRole::parse` has no such arm, and no layer emits the string: grooming is a Planner-driven flow that dispatches under the role name `planner` (`djinn-agent/src/roles/mod.rs`). It was unreachable from both callers. **Removed rather than made reachable** — making it reachable would mean inventing a dispatch role nothing produces. An existing coordinator test already asserts `Some("grooming")` is Unclassified at admission, and a new test pins that it now takes the fail-safe path.

## Tests

- `djinn-k8s` `role_classed_cpu_request_covers_every_role_and_default` extended: the launcher sidecar and `DJINN_LAUNCHER_LEASED_MILLICORES` must be **identical** for light and build-capable. The CPU *request* is the only thing the class moves — this locks role-agnostic leasing at the rendered-manifest level.
- `djinn-agent` `lease_invocation_config_carries_no_role_field`: a compile-time assertion — an exhaustive destructure with no `..` rest pattern. Adding `role` (or `resource_class`, or `is_light`) to `LeaseInvocationConfig` breaks the build; you have to delete the test to do it, and that is a visible act.
- `class_label_domain_is_exactly_two_values_on_both_carriers` (telemetry, render-level) and `class_label_vocabulary_is_bounded_to_exactly_two_values` (agent, cross-crate): the label domain is exactly two values, and the `djinn-telemetry` spellings are byte-equal to `RoleResourceClass::as_str()` — `djinn-telemetry` cannot depend on `djinn-runtime`, so this is the seam that stops them drifting.
- `class_label_passes_through_and_changes_nothing_else`: the class labels an observation and never alters kind/exit/duration.
- The bounded-label allowlist in `build_admission_epoch_disruption_tests` gained `class` with its two-value domain.

## Housekeeping

- Split `workspace.rs` (51073 of 51200 bytes) -> new `shell_exec.rs` for the timeout policy and the `finish_shell` observation seam, and `process_lease_tests.rs` -> `process_lease_shadow_tests.rs`. No `djinn:allow-oversize` added; the size guard passes at 0 unmarked oversized files.
- `cargo hakari verify` clean after adding `tracing` to `djinn-telemetry`.

## Not done, on purpose

- **Reviewer is not reclassified to `BuildCapable`.** 42% of task-runs compiling 8% of the time is not worth a slot 100% of the time on an already-oversubscribed pool.
- Board task `h1yv` is not reopened. Its scope was dispatch admission and it was correct; only its propagated justification text was wrong.
- `LIGHT_ROLE_AUDIT_REASON` in `build_admission.rs` still reads "never runs the project toolchain" — the same falsehood. Left alone to keep this PR's touch on that file to the single-line rename while another agent works there; worth a follow-up.


fernandobandeira:
Dequeued from the merge queue by a **semantic merge conflict**, not a flake — and by this PR's own new guard doing exactly its job.

`#2599` renamed `LeaseInvocationConfig::{queue_deadline_ms, launch_deadline_ms}` to `{queue_timeout, launch_timeout}` (bare `i64` millis to `Duration`) after this branch was rebased. `lease_invocation_config_carries_no_role_field` destructures that struct exhaustively with no `..` rest pattern, precisely so a field change cannot pass unnoticed — so it failed to compile against the merged tree.

Both red merge-queue lanes (`Plan Server Test Shards` and `Server sqlx Freshness`) were the *same* four compile errors; the four `Server Test` shards were fail-fast cascade. There was no independent sqlx problem — `sqlx-verify` builds the workspace before it checks the offline cache.

Rebased onto `88d73e159` (#2599) and updated the destructure to the new field names, keeping it exhaustive. Re-verified locally: 1817 tests across djinn-agent/telemetry/runtime/k8s, plus the coordinator admission/refinement/run-dir suites.

PR #2617, opened 2026-07-26
fix(goxi): give a brokered child a git trust anchor it cannot forge

fernandobandeira:
Sixth blocker in the goxi arm chain. Every brokered `git` command was going to fail the moment the launcher was armed.

A task workspace is created by the worker (uid 1000); a brokered command runs as `CHILD_UID` (1001). git compares the repository directory's **owner uid** to the process uid — group ownership and mode do not enter into it. Measured in a rendered launcher container on the production node, against a real `/workspace/wt` owned `1000:1000` and a real uid-1001 child:

```
setpriv --reuid=1001 --regid=1000 --clear-groups git -C /workspace/wt status
  fatal: detected dubious ownership in repository at '/workspace/wt'    (rc 128)
```

`job.rs` renders `GIT_CONFIG_COUNT`/`GIT_CONFIG_KEY_0`/`GIT_CONFIG_VALUE_0` = `safe.directory=*` onto the pod **specifically to fix this**, and `is_allowed_environment_key` has no `GIT_` entry, so `process_broker::child_environment` strips all three before the broker ever sees them.

## Why the obvious fix is wrong twice over

The same production probe settles both halves:

| child env | `git status` | `git clone --local` |
|---|---|---|
| (broker strips every `GIT_` var — today) | **rc 128 dubious ownership** | rc 128 |
| `GIT_CONFIG_COUNT=1 KEY_0=safe.directory VALUE_0=*` | rc 0 | **rc 128 dubious ownership** |
| `GIT_CONFIG_SYSTEM=<launcher anchor>` | rc 0 | rc 0 |

1. **Security.** `GIT_CONFIG_KEY_n`/`VALUE_n` sets *arbitrary* configuration, and `core.sshCommand`, `core.pager`, `core.editor`, `diff.external` and `credential.helper` are arbitrary command execution. A key-name allow-list is no defence at all here: `GIT_CONFIG_KEY_0` is the allowed **name** whose **value** names the dangerous key, and the forwarded environment originates in a `Command` an agent tool built.
2. **It does not work.** Command-scope config is stripped by git from the inner `git-upload-pack` of `git clone --local` (nurw, #2575) — and cloning `/mirror` is what the workspace is made of. Widening the allow-list would have taken the security hit *and* left the clone broken.

## The mechanism

`djinn_cgroup_launcher::git_trust` writes a launcher-owned system-scope config file and exports it as `GIT_CONFIG_SYSTEM`, derived and overridden on every spawn exactly like the parallelism pins.

* **SYSTEM, not GLOBAL** — `configure_private_dep_access` stores the installation token as a `url.<...>.insteadOf` rewrite that cargo/go/pnpm read back out of `$HOME/.gitconfig`. Redirecting global scope would silently break private-dependency fetches. System scope is purely additive.
* **`[include]`s the real `/etc/gitconfig`** — `GIT_CONFIG_SYSTEM` *replaces* system config rather than adding to it, so a bare anchor would drop whatever the image put there. Proven with real git against a real chained file, not by reading the generated text back.
* **Readiness gate, not a per-command error** — established in `main.rs` before the broker binds, so a launcher that cannot write its anchor fails readiness instead of taking work and then failing every git command in it (grkq).

## The security proof

The value is constrained, not just the key. `CommandSpec::validate` now applies `is_allowed_environment_entry`, and `GIT_CONFIG_SYSTEM` is admitted **only** when its value is byte-identical to the launcher's own anchor path:

* a spec naming any other file is `InvalidCommand` — refused, not silently corrected — with no leaf created and no child spawned;
* `GIT_CONFIG_COUNT` / `KEY_n` / `VALUE_n` / `NOSYSTEM` / `GLOBAL` / `GIT_SSH_COMMAND` / `GIT_PAGER` / `GIT_EXTERNAL_DIFF` stay wholly unforwardable, by key **and** as entries;
* `GIT_CONFIG_NOSYSTEM`'s exclusion is load-bearing rather than tidy: the suite makes real git ignore the anchor with it set and reproduce the exact production failure;
* behaviourally, with a hostile `core.sshCommand`/`core.pager` config file sitting readable in the workspace, a brokered child resolves both keys to empty and the marker the hostile config would have created never appears;
* the anchor itself is verified to grant `safe.directory` and nothing else, by asking real git what the whole SYSTEM scope resolves to.

The child cannot tamper with the file either — dir `0755` / file `0644`, owned by the launcher uid, re-verified by `lstat` (never following a symlink) on every spawn and failing closed. Measured in the launcher container: `touch` and `>>` both return `Permission denied`.

### One thing the probe found that no assertion would have

An `emptyDir` mounted at `/tmp` arrives **`2777` — world-writable with no sticky bit**, so a child could `mv` the root-owned anchor directory out of the way (rc 0). That is a denial rather than an escalation (the ownership check fails closed on a displaced directory, and both children share a uid anyway), but it is free to remove: the launcher restores `1777` semantics on a scratch root it owns. Re-measured on the node, the same `mv` then returns `Operation not permitted`.

## The umask question

**The brokered path does not depend on an ambient umask.** The worker pins `0002` itself at startup (`volume_contract::apply_artifact_umask`, called before any filesystem work) and `child::prepare_child` pins `0002` for the child. What was missing is a test pinning what that buys, so the privileged proof measures it on real uids:

```
umask 0002 -> worktree 2775 -> uid 1001/gid 1000 writes it        (rc 0)
umask 0022 -> worktree 2755 -> uid 1001/gid 1000 Permission denied (rc 1)
```

## Testing

Nothing here asserts that a key appears in an env map.

**Always-on** (`tests/brokered_git_trust.rs`, runs in every shard): builds a real repository, takes the environment out of the real `Launcher::create_command` through a real cgroup/spawn composition, and runs real `git`. It reproduces `detected dubious ownership` **by name** with the fix reverted — the reverted case is literally `job.rs`'s rendered env put through `is_allowed_environment_key`, so it is the production configuration and not a mock of it. `GIT_TEST_ASSUME_DIFFERENT_OWNER` (git's own hook into `ensure_valid_ownership`) makes the check fire as any uid, so these fail rather than skip.

**Privileged** (`launcher-kernel-boundary` lane, `#[ignore]`d): the genuinely cross-uid version — a worktree chowned to the worker uid, and `git status` run by a process that has `setgroups(0)`/`setresgid`/`setresuid`'d to `CHILD_UID` in the same order `prepare_child` uses, with **no test hook** — plus the umask measurement. An always-on guard reads the workflow back and fails if the lane, its `--ignored` invocation, or the `GIT_TRUST_EXPECTED_PROOFS` declaration disappears, so a run that executes zero proofs can never look green. Both were executed locally in a container as uid 0 and pass.

**Production**: validated on `13.140.147.151` with a probe pod rendered as the launcher container (`readOnlyRootFilesystem`, uid 0, `/tmp`+`$HOME`+workspace emptyDirs) — never armed. Every row of the table above, the tamper attempts, the sticky repair, and the umask arithmetic are that probe's output.

## Seventh blocker

None found in this pass. The nearest remaining risk is documented in the review notes rather than silently left: `$HOME` (`/home/djinn`) is now a writable emptyDir shared by every brokered child, so `git config --global` from one child is visible to the next. That is not new to this change and not a privilege boundary (all children share uid 1001), but it is the surface an armed run will exercise first.

## Dependencies

Rebased onto #2616, which supplies the writable `/tmp` this depends on. Without it the readiness gate would fail closed rather than fail silently — which is the intended direction, but it does mean these must land in this order.

### DIFF
--- a/.github/workflows/quality-gate.yml
+++ b/.github/workflows/quality-gate.yml
@@ -562,6 +562,15 @@ jobs:
       # fewer than this many actually execute, so a lane that quietly runs zero
       # tests can never be mistaken for a proven kernel boundary.
       ZF13_EXPECTED_PROOFS: "4"
+      # Number of `#[ignore]`d proofs `brokered_git_trust.rs` declares. They are
+      # the only place the cross-UID git contract is proven for real — a
+      # worktree owned by the WORKER uid and `git status` run by a process that
+      # has `setresuid`'d to CHILD_UID, with no test hook standing in for the
+      # uid mismatch. The suite's always-on
+      # `the_privileged_git_proofs_are_wired_and_cannot_silently_skip` reads
+      # this file back and fails if this declaration or the run below
+      # disappears (task goxi, sixth launcher blocker).
+      GIT_TRUST_EXPECTED_PROOFS: "2"
       # Number of `#[ignore]`d proofs `delegated_cpu_lease_lifecycle.rs`
       # declares. That suite is the ONLY place the CPU lease is proven to
       # actually govern CPU — measured on `cpu.stat` over a wall-clock window
@@ -613,6 +622,7 @@ jobs:
           cargo test -p djinn-cgroup-launcher \
             --test kernel_boundary_under_rendered_context \
             --test delegated_cpu_lease_lifecycle \
+            --test brokered_git_trust \
             --no-run --message-format=json > "$RUNNER_TEMP/build.json"
           binary=$(jq -r 'select(.executable != null and .target.name == "kernel_boundary_under_rendered_context") | .executable' "$RUNNER_TEMP/build.json" | tail -1)
           if [ -z "$binary" ] || [ ! -x "$binary" ]; then
@@ -626,6 +636,12 @@ jobs:
             exit 1
           fi
           echo "lease=$lease" >> "$GITHUB_OUTPUT"
+          git_trust=$(jq -r 'select(.executable != null and .target.name == "brokered_git_trust") | .executable' "$RUNNER_TEMP/build.json" | tail -1)
+          if [ -z "$git_trust" ] || [ ! -x "$git_trust" ]; then
+            echo "::error::could not locate the built brokered git-trust proof binary"
+            exit 1
+          fi
+          echo "git_trust=$git_trust" >> "$GITHUB_OUTPUT"
       - name: Prove the kernel boundary under the rendered security context
         run: |
           set -euo pipefail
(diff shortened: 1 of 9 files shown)


fernandobandeira:
### CI caught the suite lying — fixed in `cd4503946`

Shard-4's failure was the non-vacuity control working exactly as intended:

```
the reverted environment must reproduce `detected dubious ownership`;
got status=ExitStatus(unix_wait_status(0)) stderr=""
```

A GitHub Actions runner ships `/etc/gitconfig` containing `[safe] directory = *` — machine-wide trust for every repository and every process. The harness already cleared the environment, so `$HOME/.gitconfig` was out of the way, but **SYSTEM scope is not environment** and came through untouched. With it, the non-vacuity control passes while the blocker is fully present, and the positive control passes for a reason the launcher had no part in. Both arms were vacuous on that runner; only the negative one had the decency to say so.

The fix excludes the **host**, not the mechanism. `git_trust::materialize_in_with_system` writes the same bytes through the same production writer but chains to a caller-supplied file, and the fixture gives every behavioural arm a controlled empty system config. That is the faithful model rather than a convenience: measured inside a rendered launcher container on the production node, `/etc/gitconfig` **does not exist**, so the only protected-scope configuration a brokered child can ever see is the one the launcher hands it.

Two things got stronger as a result:

* the arms now differ in exactly **one** variable — empty file vs anchor — instead of differing in "anchor present" against an unknown host baseline;
* `the_anchor_grants_safe_directory_and_no_other_protected_setting` can now assert SYSTEM scope holds the trust rule and *nothing else*, which was impossible while the host was contributing entries.

The composition claim moved to its own proof, `the_brokered_child_environment_names_the_launchers_own_anchor`: the environment out of the real `Launcher::create_command` carries exactly one git variable and it is `anchor_path()`. That half needs no git run, so it is not entangled with the host filesystem at all.

Verified by **reproducing the CI condition** rather than reasoning about it — with `[safe] directory = *` planted in a container's `/etc/gitconfig`, the suite is 9/9 always-on green and 2/2 privileged green, where the previous revision went vacuously green on the same host.

PR #2656, opened 2026-07-27
fix(cargo-target-seed): copy build-script output instead of hardlinking it

fernandobandeira:
## Why

A hardlinked build-script `OUT_DIR` payload cannot be rewritten when cargo re-runs the build script.

`fs::copy` finishes by applying the source's mode to the destination — an `fchmod` on the destination fd — and `fchmod` is **EPERM for a file the caller does not own, regardless of mode**. A hardlinked entry keeps the warm base's owner, so the copy fails `Operation not permitted` even though the bytes were writable through the shared gid. Note EPERM, not EACCES: this is an *ownership* failure, which is why `ensure_owner_writable` cannot rescue it.

`libssh2-sys/build.rs` does exactly this and `.unwrap()`s it:

```rust
54: fs::copy("libssh2/include/libssh2.h", include.join("libssh2.h")).unwrap();
```

So the panic killed the compile before it reached any workspace crate. Task-run pods whose build-script fingerprint went stale could never build locally and fell back to guessing against CI at ~20 min per attempt — task `gy53` burned **51 sessions over 10 hours** this way, with both work submissions reporting the same blocker.

The sharing is a corruption hazard in the other direction too: where a rewrite *does* land, the shared inode writes **through** into the per-project warm base that every other task-run pod seeds from.

This is the identical failure already carved out for `.rustc_info.json` and the cargo lock files. Build-script state is the third instance, and it was in the default `Hardlink` arm only by omission.

## What

- Classify `build/<pkg>-<hash>/out/**` and the in-place stamp files (`output`, `stderr`, `root-output`, `invoked.timestamp`) as `Copy`. The destination is then owned by the seeding process — which is also the process that later runs cargo.
- `build-script-build` executables **stay hardlinked**: cargo replaces them by rename, never in place, and they are the heavy part of `build/`.
- Matching is positional (`build/*/out`), not "has a component named `out`", so unrelated artifacts are not swept in and silently converted from hardlinks into byte copies.

## Test fallout worth reviewing

The foreign-owner fixture used build-script `OUT_DIR` payloads as its **only** unlinkable entries, so this reclassification would have silently retired the link-fallback coverage entirely. It gains ordinary compiled artifacts at `0644`/`0444`/`0755` — the shape a base built under `umask 022` produces — so that path is still exercised on its own merits.

The symlink fixture moves out of a build-script dir for the same reason: under the new classification its `CloneAction::Copy` assertion would have passed even with symlink handling removed.

## Verification

`cargo test -p djinn-agent-worker --lib` — 46 passed.

Verified by **neutralization**: reverting the classification arm alone fails the new tests, including the load-bearing side-effect test that rewrites a seeded payload and asserts the warm base is byte-for-byte unchanged.

## Tradeoff

Build-script outputs are now byte-copied per run rather than shared by inode, which costs private disk in `/cache/cargo-target-runs` (a path with a history of node DiskPressure). Copying is the cheaper of the two correct options — the alternative, `Skip`, would force every vendored `-sys` crate to recompile from scratch on every task run.

### DIFF
--- a/server/crates/djinn-agent-worker/src/cargo_target_seed.rs
+++ b/server/crates/djinn-agent-worker/src/cargo_target_seed.rs
@@ -6,6 +6,21 @@
 //! dir. Mutable Cargo metadata is copied, and incremental state is skipped so it
 //! is never shared by inode with the warm base.
 //!
+//! # Only immutable payloads may be hardlinked
+//!
+//! The rule that governs [`classify_cargo_target_path`] is: **anything the run
+//! may rewrite must be copied, not linked.** A hardlink shares the inode with the
+//! warm base, which breaks such a path two ways — the rewrite either fails
+//! outright (`fchmod` on a file the caller does not own is EPERM regardless of
+//! mode, so `fs::copy` into a linked destination reports "Operation not
+//! permitted") or succeeds and writes THROUGH into the base that every other
+//! task-run pod seeds from.
+//!
+//! Four classes are known to be rewritten in place and are carved out for that
+//! reason: cargo's per-profile lock files, `.rustc_info.json`, build-script
+//! `OUT_DIR` payloads, and the build-script stamp files. Adding a fifth means
+//! adding it here — the default arm is `Hardlink`, so an omission is silent.
+//!
 //! # Why a single entry must never discard the base
 //!
 //! The warm base is written by the warm Job (uid 10001) and read by the task-run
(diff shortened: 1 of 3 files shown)


fernandobandeira:
**Converting to draft — the stated premise of this PR is false in production.**

This PR justifies `Copy` with: *"a byte copy makes the destination owned by the seeding process — which is also the process that later runs cargo."* Measured on the production cluster, those are three different uids:

| stage | uid | evidence |
|---|---|---|
| warm base | **10001** | `find /…/djinn-cache/cargo-target -maxdepth 3 -printf '%U'` → `10001` |
| seeder (worker container) | **1000** | pod spec `containers[worker].securityContext.runAsUser = 1000` |
| cargo / build scripts | **1001** | `cgroup-launcher` sidecar renders and runs (initContainer, `restartPolicy: Always`, uid 0, 10 min uptime, 0 restarts); it `setresuid`s children to `CHILD_UID = 1001` |

So `Copy` moves the destination owner from 10001 to 1000, but the build script is uid 1001 either way — `fs::copy`'s terminal `fchmod` still returns EPERM. **This does not fix the reported failure.**

Note the in-repo comment at `job.rs:525-529` says `validate_enforcement_render` "rejects the armed mode outright, so in practice no task-run pod renders it". That comment is **stale** — the function validates broker quota, cgroup profile readiness and volume ownership rather than rejecting unconditionally, and live task-run pods do render the sidecar.

Second, independent problem with this change: `fs::copy` does not preserve mtime but hardlinks do, so moving build-script stamps and `out/**` to `Copy` gives them seed-time mtimes while `.rlib`s keep warm-time mtimes. Cargo's `StaleDependency` check is a raw mtime comparison, so this can bust fingerprints that were previously `Fresh` — trading one EPERM for a broader, ordering-dependent rebuild.

Holding until both are addressed: the seed must be owned by (or chowned to) `CHILD_UID`, and copied entries must preserve source mtimes.

fernandobandeira:
Closing in favour of a correct fix. The classification change here (build-script `out/` must not be hardlinked) is still necessary — the write-through corruption hazard is real regardless of uid — but it is not sufficient, and shipping it alone would trade one failure for an mtime-driven one. Superseded by the ownership work; the classification change will be carried over there with the mtime fix.

PR #2695, opened 2026-07-28
feat(warm): release the build lease at the cargo→graph boundary (~20% of build capacity)

fernandobandeira:
MEASURED from `build_leases` over 6h48m on 2026-07-27.

## The waste

| measurement | value |
|---|---|
| warm Job's hold on 1 of only 3 build slots | **6h44m of 6h48m — 98.9% duty cycle** |
| share of that hold spent in the SCIP/graph phase | **60.7%** |
| share of the whole Job that is ONE rust-analyzer process | 48.5% |
| that process's average CPU | **~0.82 cores** (2899 s CPU / 3520 s elapsed, rigorously bounded; the pod's cgroup allows 4) |
| the lease's weight | 4000 millicores |

The cargo phase genuinely is a build and genuinely needs that weight. The graph phase is not a build — it is one mostly-serial indexer plus serialization and publication — and it was charging a full build-capable slot for ~0.82 cores of single-threaded work.

The warm now hands its slot back the moment `warm_cargo_target_base` returns, **before** `run_warm_graph_command`. This recovers ~0.6 of 3 slots — **~20% of total build capacity** — with **no change to graph freshness**: the graph phase runs exactly as before, it just stops charging build capacity for time it does not use.

## How

`build_leased_warm_job` projects the durable lease identity into the warmer container:

- `DJINN_WARM_LEASE_CONSUMER_ID` — the lease's `warm_request_id`
- `DJINN_WARM_LEASE_FENCING_TOKEN`

Only the **leased** path projects them; an unleased warm holds no slot and must not attempt a release (asserted both ways).

`warm_cargo_and_continue` grows an explicit boundary hook between the cargo phase and the graph continuation, and the worker releases there via the new `warm_build_lease` module.

**This is not a second lease authority.** Queue, grant and bind stay host-owned in `djinn-coordinator`. The Pod does exactly one thing: a fenced, idempotent release of a slot it already holds, through the same `BuildLeaseRepository::release` primitive the coordinator uses, against the same row.

## Exactly-once, and leak-free

Three independent layers, because "release twice" and "leak the slot" are both production incidents:

1. **In-process** — `release_once` latches an `AtomicBool` with `swap`. A second call on the same handle does no database work at all.
2. **Durable** — `release` is a compare-and-set on the fencing token inside a locked transaction, and `BuildLeaseRepository::terminal` **replays** an already-terminal row rather than re-terminalising it (it does not even re-stamp `terminal_at`). The host's own release at Job termination, the reclaim sweep, and a restarted coordinator's recovery therefore all land on an idempotent no-op. A stale token is rejected outright, so a Pod outlived by a newer grant can never free somebody else's slot.
3. **Crash-safe** — a crash *after* the release cannot leak the slot: the row is already terminal, and terminal rows are excluded from occupancy, so the FIFO keeps draining with no reclaim sweep, no settle window and no operator. A crash *before* the release leaves behaviour exactly as it is today — the host-side absence proofs in `djinn_coordinator::build_lease_reclaim` and `K8sGraphWarmer::reconcile_durable_warm_leases` still own that row, and this change touches neither.

A failed release is also safe: it is logged and the warm proceeds. The only cost of a failure is the capacity win, never correctness.

## Interaction with #2688

#2688 (coordinator-side: release on terminal warm workload rather than object absence) is complementary and there is no conflict — that governs the **host's** release when the Job ends, this governs the **in-pod phase boundary**. Whichever runs second is a no-op replay: `terminal()` returns the retained result for an already-terminal row without mutating, and `recoverable()` filters terminal rows, so once the in-pod release terminalises the row that reconciler pass never sees it.

## Not the final answer

A spike measured that splitting SCIP into its **own** Job (request 1 CPU, limit 2, 16Gi, no build-slot weight) is strictly stronger: rust-analyzer's SCIP pass is single-threaded by design (upstream rust-analyzer#18140, open since 2024), so the 4-core reservation is structurally wasted, not merely oversized. This change composes with that and is worth landing now; it is not a reason to stop there.

## Tests — every assertion is a capacity side effect

Never a state column read back; the question asked throughout is "is a slot actually grantable to somebody else".

| test | asserts |
|---|---|
| `releasing_at_the_boundary_frees_a_slot_another_consumer_can_take` | a second consumer is **refused at cap** while the warm holds it, and **GRANTED** after the release |
| `release_is_exactly_once_in_pod_and_idempotent_against_the_host` | the second in-pod call does no DB work; the host's later release replays without re-stamping `terminal_at` |
| `a_second_handle_reports_a_no_op_rather_than_a_second_freed_slot` | bypassing the latch still cannot book a second freed slot |
| `a_crash_after_the_release_leaks_nothing` | the waiting consumer is granted with **no** host intervention |
| `a_crash_before_the_release_leaves_the_slot_to_the_host_reclaim` | the slot is **NOT** grantable — so a regression that released early is caught here |
| `a_stale_fencing_token_cannot_release_the_slot` | a fenced-out release leaves the row occupying |
| `the_build_slot_is_released_before_the_graph_phase_not_after_it` | the **ORDER** — releasing after the graph phase is indistinguishable from the old behaviour |
| `the_leased_warm_job_renders_the_build_lease_identity_the_worker_releases_with` | joins the **real** rendered manifest to the **real** worker constants in `rendered_job_env_contract` |
| `the_unleased_warm_job_renders_no_build_lease_identity` | the other direction |

The contract test matters more than it looks: `WarmBuildLease::from_env` deliberately returns `None` for anything it cannot prove, so a renamed variable would not crash the warm — it would silently give back the entire ~20% capacity win, on every warm, forever. That is precisely the failure mode `rendered_job_env_contract` exists to catch.

### Neutralization results

Every new test was verified by reverting the production change and confirming failure:

| neutralization | failing tests |
|---|---|
| `release_once` returns `Released` without touching the ledger | **6 of 8** |
| the leased manifest stops projecting the lease env | **1** (the contract test) |
| the boundary hook moves to *after* the graph phase | **1** (the ordering test) — and note this one initially passed, which is why the explicit ordering assertion was added rather than relying on "the hook ran" |

`cargo fmt --all --check` clean, `scripts/check-file-size.sh` reports 0 oversized files, and `cargo clippy` adds no warnings in the touched files (the two `Instant::now()` errors in `crates/djinn-k8s/src/graph_warmer_ledger_tests.rs` are pre-existing and owned elsewhere).


fernandobandeira:
Dequeued briefly to push one test-only fix (1beebf4): the new boundary-ordering test armed the process-global `WARM_CARGO_INJECTED_LOCK_FAILURE` without holding `WARM_CARGO_ORDERING_MUTEX`, so under a contended run it could steal a sibling test's lock acquisition and cascade through the three `warm_cargo_ordering_*` tests via a poisoned mutex. That is exactly the shape of flake that wedges the merge queue for everyone, hence the dequeue rather than letting it ride. Re-queueing once checks are green.

PR #2697, opened 2026-07-28
feat(scip): split the semantic index into its own leaseless Kubernetes Job

fernandobandeira:
Splits the SCIP (semantic) indexing phase out of the canonical-graph warm Job into its own Kubernetes Job that **acquires no build lease**.

**The benefit is conditional. Read §6 before deciding to arm this.** It is a large win on a repository whose `main` is quiet for an hour at a time, and close to zero on one that merges more often than hourly — which includes djinn's own repo on a busy day. The feature therefore ships **disarmed** (`DJINN_K8S_SCIP_INDEX_ENABLED`, default off) with a quiescence gate that declines to index a tree that is still moving.

---

## 1. The problem, and why resizing cannot fix it

A complete production warm measured **90m42s** end to end, of which the SCIP index alone was **58m43s (~65%)**.

That phase is *structurally* single-threaded. rust-analyzer's `StaticIndex::compute` is a bare `for module in work` loop with no rayon; the measured parallel window was **20s against a 368s serial window (94% serial)**. Upstream `rust-lang/rust-analyzer#18140` is open and maintainer-confirmed. So the warm Job reserved a full build slot sized for a 4-core parallel cargo build, then spent roughly two thirds of its life using one core.

**The slot arithmetic.** `BuildSlotWeight::for_millicores` (`server/crates/djinn-runtime/src/spec.rs`) is:

```rust
let slots = millicores.div_ceil(slot_millicores);
Self(i64::from(slots.max(1)))
```

`div_ceil` then `.max(1)`. **Any nonzero CPU request costs at least one whole build slot.** Lowering the warm request from 2 CPU to 500m changes the charged weight not at all. And `build_lease.rs` resolves `LeaseIdentity::GraphWarm(_) => self.weights.warm()` — warm is its own lease identity, charged as a unit.

The only lever that returns capacity is for the SCIP work **not to acquire a build lease at all**, which means it cannot be a `GraphWarm` consumer, which means it must be its own Job. That is what this PR builds.

## 2. Proposal 8ixk: the protected label and the ≤2200m budget

A Job that acquires no lease **and** is not labelled protected is *invisible CPU* — it burns real cores that no cap can account for. That is precisely the oversubscription proposal `8ixk` ("Node-keyed build slot pools") exists to prevent. 8ixk derives each node's cap as:

```
floor((allocatable_mcpu - protected_mcpu) / cores_per_slot_mcpu)     cores_per_slot = 2800
```

where `protected_mcpu` sums the CPU requests of Pods carrying `djinn.io/capacity-reserved=true`.

So the SCIP Job carries that label — on the Job **and** on the Pod template, since 8ixk selects Pods. On the current 12-core VPS with 4200m already protected (server 2000m + postgres 1000m + buildkitd 1000m + qdrant 100m + zot 100m):

| | derivation | cap |
|---|---|---|
| today | `floor((12000 − 4200) / 2800)` = `floor(2.78)` | **2** |
| with this Job @ 1000m | `floor((12000 − 5200) / 2800)` = `floor(2.43)` | **2** (unchanged) |

### ⚠️ The request budget is 2200m. Do not raise it past that.

Solved from the same formula — the cap holds at 2 while `12000 − 4200 − request ≥ 2 × 2800`, i.e. while `request ≤ 2200`:

| request | derived cap |
|---|---|
| 2200m | 2 |
| 2201m | **1** |

Raising this Job's CPU request past **2200m** costs the deployment a whole build slot — half its build concurrency under 8ixk. This is not a sentence in a PR description that rots: it is `SCIP_PROTECTED_REQUEST_CEILING_MILLICORES`, and `scip_cpu_request_stays_inside_the_8ixk_protected_budget` proves it by **recomputing 8ixk's formula from its inputs**, including asserting that 2201m provably drops the cap. A future edit that raises the request fails the build.

*(Note this PR does not implement 8ixk, which is still in draft. It makes the new Job forward-compatible with it — the label is inert until 8ixk lands, and the arithmetic is asserted now so it cannot silently stop holding.)*

## 3. The 16Gi memory floor and its regression test

Measured peak RSS for the SCIP phase is **10.0 GB**.

`server/crates/djinn-k8s/src/config.rs` has `warm_memory_limit` defaulting to `"6Gi"` — **below the measured peak**. Production survives only because an out-of-band override raises it to 16Gi. That arrangement (a default that would OOM, rescued by an override nothing tests) is exactly what this PR refuses to repeat.

`scip_memory_limit` is `"16Gi"` explicitly, and `scip_memory_limit_covers_the_measured_peak` fails the build if the rendered limit ever drops below `MEASURED_SCIP_PEAK_MEMORY_BYTES`. It also asserts that `6Gi <` the measured peak — so the test cannot pass for a Pod that OOMs, which is the property that makes it a real regression test rather than a tautology.

## 4. Scheduling and the change-detection gate

`server/crates/djinn-k8s/src/scip_schedule.rs`. Every 3 hours, **and only if the head advanced**.

```
head unknown            → SkipUnknownRevision      (fail closed)
inventory unreadable    → SkipInventoryUnavailable (fail closed)
non-terminal Job exists → SkipInFlight
last indexed == head    → SkipHeadUnchanged     ← the change-detection gate
head younger than ~59m  → SkipHeadNotQuiescent   ← the hit-rate gate (§6)
elapsed < 3h            → SkipCadence           ← the cadence floor
otherwise               → Dispatch
```

Ordering is load-bearing: **head-unchanged is checked before the cadence**, because the requirement is "every 3 hours *and only if* the head advanced", not "or". `unchanged_head_never_dispatches_however_much_time_passed` asserts this across elapsed times from zero to 30 days.

**It does not use the existing warm debounce, which is dead.** `graph_warmer.rs` has `DEFAULT_WARM_DEBOUNCE_SECONDS = 180` and `DEFAULT_WARM_DEBOUNCE_MAX_WAIT_SECONDS = 900`, but `server/src/mirror_fetcher.rs` (`DEFAULT_INTERVAL_SECS = 60`) calls `graph_warmer().trigger(project_id)` unconditionally every 60s for every project with indexable code. The 180s quiet window can therefore never elapse; every window runs to the 900s cap — verified in production, a window opened and collapsed at exactly 900.000s having coalesced 17 triggers. This gate is **pull, not push**: it is driven by its own tick and reads observed state. It never subscribes to `trigger`.

**Where the "last successful index" state lives — and why there is no migration.** Each SCIP Job is named deterministically from `(project, revision)` and annotated with `djinn.app/scip-revision`; `scip_job_ttl_seconds` (4h) is deliberately longer than `scip_index_interval_seconds` (3h). The retained *succeeded* Job set **is** the ledger, and it is durable across server restarts because it is apiserver state. `scip_job_ttl_outlives_the_index_interval` pins the relationship.

The failure mode of an early-GC'd Job is a redundant dispatch, and a redundant dispatch is cheap: the SCIP artifact cache is content-addressed, so re-running against an unchanged tree hits the cache for every workspace. Losing the ledger costs a fast no-op, never a wrong graph. That is worth avoiding a schema migration (and its golden fanout) for.

Only a **succeeded** Job advances the ledger. A failed index leaves the last known-good revision in place, so the head still reads as advanced and a later tick retries — gated by the cadence floor, so a crash-looping index cannot re-create a leaseless 16Gi Pod every tick. Both halves are asserted separately.

## 5. Failure / skip degradation — the graph never degrades to empty

**The SCIP Job never publishes a graph.** `run_scip_index_command` writes no `repo_graph_cache`, `repo_graph_generation`, or `repo_graph_current` row. It only fills the content-addressed SCIP cache.

So there is no code path from "SCIP Job failed / was skipped / never ran" to a degraded graph. When it does not run, `ensure_canonical_graph` re-runs the indexers inline **exactly as it does today**, still under:

* the `node_count == 0` early return that refuses to write an empty blob, and
* per-workspace salvage (`graph.salvage_workspace_from_artifact`) that splices the previous artifact for any workspace whose indexer failed.

The worst case is a **slower warm**, never an empty graph. Every non-`Dispatch` outcome including every error is a skip, and uncertainty resolves toward not dispatching, because dispatching on unknown state is the only branch that can cost anything.

### The mechanism

The SCIP indexers already write through `ScipCacheStore`, whose root resolves from `$XDG_CACHE_HOME` — which `job.rs::cache_env_vars` points at the shared `/cache` PVC (`/cache/xdg/<project>`). Its `WorkspaceIdentity` holds a **relative** workspace root, so an artifact produced from the SCIP Pod's clone is a `CachedHit` for the warm Job's `execute_plan_with_cache` even though the two Pods cloned to different absolute paths. **No change to `ensure_canonical_graph` is required, and none is made.**

### ⚠️ The win is conditional, and here is exactly when it does not materialise

`CacheKeyIngredients` includes `source_hashes` over the workspace's sources. So an index produced at head `H0` is only a hit for a warm running against **`H0`'s sources**.

Warm chases head continuously: `mirror_fetcher` triggers every 60s, and the debounce is dead (§4), so every window collapses at its 900s cap — a warm roughly every 15 minutes. The SCIP Job takes ~59 minutes. Therefore:

| condition | outcome |
|---|---|
| `main` quiet across the SCIP run | warm hits the cache; its SCIP phase collapses to seconds; **slot-time 90m42s → ~32m (−65%)** |
| `main` advances during the SCIP run | warm **misses** and re-indexes inline for the full 58m43s; slot-time unchanged; the SCIP Job's ~59 min of CPU was **redundant** |

**On a repository merging more often than once an hour the hit rate tends toward zero.** That plausibly describes djinn's own repo on an active day. This is not a caveat about measurement error — it is a structural property of a content-keyed cache feeding a consumer that chases head.

Two things bound the downside of a miss:

* The redundant work costs **no build slot** — the SCIP Job is leaseless, and its 1000m request is accounted as protected capacity (§2). It cannot starve dispatch; it consumes one core of the node's protected reservation.
* A miss is not a *regression*. The warm behaves exactly as it does today.

The key is also **per-workspace**, so a commit touching only workspace A leaves workspace B's key intact and B still hits. That genuinely helps a polyglot repo — but not the case that dominates the cost here, since djinn's Rust `server/` workspace is both the most expensive to index and the most frequently touched.

### The mitigation actually implemented: the quiescence gate

`decide` will not dispatch unless the head has already stood still for `scip_quiescence_seconds`, defaulting to **3523s — the measured phase cost itself**. The rule is: *do not index a tree that has not been stable for at least as long as indexing it takes.* That is the condition under which the artifact is likely to still be current when it lands.

It cannot prevent a commit arriving *during* the run — nothing can. What it does is stop the system from spending 16Gi and an hour of a core producing an artifact that was already provably unlikely to be used. An unreadable commit time reads as not-quiescent, consistent with every other uncertainty here.

**What would make the win unconditional** is letting the graph build consume a *stale but present* semantic index, so the SCIP cadence could diverge from the warm cadence freely. That is a real architectural change to the publication path (see the last section) and is not attempted here.

### Syntactic and git-derived passes stay on the warm path

They are deliberately not moved. `ingest_coupling_best_effort` (already hoisted above the cache check and the indexer lock, so it is independent of SCIP), the tree-sitter post-processors (`entry_points`, `processes`, complexity, `db_access`, route extraction) and `cochange` want per-commit freshness and are cheap. They continue to run on the warm path at its own cadence. Only the expensive semantic pass moves to the 3-hour cadence.

## 7. Cache/compile parity — derived from `warm_cpu_limit`, not this Pod's own limit

This looks wrong and is not, so it is asserted rather than commented:

* `CARGO_BUILD_RUSTFLAGS` embeds `-Clink-arg=-Wl,--threads=N` derived from the CPU limit. **RUSTFLAGS feed rustc's `-C metadata`**, so a `--threads=2` render would give every compilation unit a different fingerprint from the warm base and discard 100% of it.
* `CARGO_TARGET_DIR` is keyed `.../mold-jobs-N` from the same limit — a divergent N forks a second, cold target base on the PVC.
* `CARGO_INCREMENTAL=1` / `RUSTC_WRAPPER=""` are the shared compile strategy every djinn build pod must agree on (`job.rs` ~line 1055 explains why); cargo folds both into its fingerprints.

`scip_job_cache_env_matches_the_warm_job_render` sets `warm_cpu_limit = "6"` and `scip_cpu_limit = "2"` — deliberately different — and asserts the SCIP Job's rendered env equals the **warm Job's actual rendered env** for all eight cache-relevant vars. A render that keyed off `scip_cpu_limit` fails it.

## 8. Relationship to #2695

Orthogonal and composable; **no file overlap**.

* **#2695** releases the warm Job's build lease at the cargo→graph boundary *inside the same Pod* — shortening how long the warm holds its slot.
* **This PR** removes the SCIP phase from the warm Pod entirely — so there is no 58m43s SCIP phase left inside that hold to release.

This diff does not touch `warm_job.rs` at all (`scip_job.rs` only reads its public constants), and touches `graph_warmer.rs` only to add one 25-line accessor. Rebasing either PR against the other should be mechanical.

---

## Changes

| File | |
|---|---|
| `djinn-k8s/src/scip_job.rs` | **new** — Job manifest, labels, 8ixk arithmetic constants, 10 tests |
| `djinn-k8s/src/scip_schedule.rs` | **new** — cadence + change-detection gate, inventory projection, 12 tests |
| `djinn-k8s/src/config.rs` | `scip_*` knobs + `DJINN_K8S_SCIP_*` env parsing + doc table |
| `djinn-k8s/src/lib.rs` | module registration / re-exports |
| `djinn-graph/src/canonical_graph.rs` | `run_scip_index_command` — the semantic pass alone |
| `djinn-agent-worker/src/main.rs` | `scip-index` subcommand |
| `djinn-agent-worker/src/rendered_job_env_contract.rs` | the SCIP Pod's argv parses with the real clap parser |
| `djinn-k8s/src/graph_warmer.rs` | `scip_index_scheduler()` accessor (additive, 25 lines) |
| `djinn-server/src/scip_index_watcher.rs` | **new** — the leader-only tick |
| `djinn-server/src/server/state/mod.rs` | spawn it in `become_leader` |

2627 insertions, 19 deletions. `warm_job.rs` is untouched.

### Configuration

| Env | Field | Default |
|---|---|---|
| `DJINN_K8S_SCIP_CPU_REQUEST` | `scip_cpu_request` | `1` — **≤ 2200m** |
| `DJINN_K8S_SCIP_CPU_LIMIT` | `scip_cpu_limit` | `2` |
| `DJINN_K8S_SCIP_MEMORY_REQUEST` | `scip_memory_request` | `4Gi` |
| `DJINN_K8S_SCIP_MEMORY_LIMIT` | `scip_memory_limit` | `16Gi` — measured peak 10.0 GB |
| `DJINN_K8S_SCIP_INDEX_INTERVAL_SECONDS` | `scip_index_interval_seconds` | `10800` (3h) |
| `DJINN_K8S_SCIP_JOB_TTL_SECONDS` | `scip_job_ttl_seconds` | `14400` (4h, > interval) |
| `DJINN_K8S_SCIP_JOB_TIMEOUT_SECONDS` | `scip_job_timeout_seconds` | `7200` (measured phase 3523s) |
| `DJINN_K8S_SCIP_INDEX_ENABLED` | `scip_index_enabled` | **`false`** — the arming switch |
| `DJINN_K8S_SCIP_QUIESCENCE_SECONDS` | `scip_quiescence_seconds` | `3523` — the measured phase cost; `0` disables the gate |
| `DJINN_SCIP_INDEX_TICK_SECS` | watcher loop period | `300` |

No `EnvironmentConfig`, build-resource shape, MCP tool schema, or migration was touched, so there is no golden/snapshot fanout to regenerate.

## Testing

```
cargo test -p djinn-k8s   --lib                     286 passed
cargo test -p djinn-graph --lib                     662 passed
cargo test -p djinn-server --lib scip_index_watcher   2 passed
cargo test -p djinn-agent-worker --bin ...          rendered_job_env_contract: 3 passed
cargo fmt --all -- --check                          clean
cargo clippy -p djinn-k8s -p djinn-server -p djinn-graph -p djinn-agent-worker --all-targets
                                                    no new warnings
```

## Activation

The scheduler is wired **for real** at the composition root — `scip_index_watcher::spawn` sits next to `mirror_fetcher::spawn` in `become_leader`, resolves the cluster-backed `ScipIndexScheduler` from the live warmer (sharing its `kube::Client` rather than building a second one), and calls `tick_project` per project.

**Whether it creates Jobs is decided inside the tick**, by `DJINN_K8S_SCIP_INDEX_ENABLED` (default `false`). Gating the *wiring* rather than the *action* is how a feature ends up merged and permanently unreachable, so:

* arming is a config flip, not a redeploy;
* a disarmed deployment still resolves the real scheduler every tick;
* if the flag is armed and no scheduler resolves, the watcher logs at **ERROR** — a wiring regression is loud, not silent.

Suggested rollout: leave it off through this deploy, then set `DJINN_K8S_SCIP_INDEX_ENABLED=true` and watch for `scip_index_watcher: dispatched standalone semantic index` plus the resulting Job's outcome. The decision reason is on every tick at debug (`head_unchanged`, `head_not_quiescent`, `cadence`, `in_flight`, …), so it is possible to see *why* nothing is dispatching before concluding it is broken.

## What is deliberately not in this PR

**Letting the graph tolerate a stale semantic index.**  letting the graph build consume a *stale but present* semantic index (so the SCIP cadence could diverge freely from the warm cadence). `ensure_canonical_graph` currently builds every node and edge from `&[ParsedScipIndex]` — the syntactic passes are post-processors on a SCIP-built graph, not independent producers. Making the graph tolerate a stale index is a genuine architectural change to the publication path and belongs in its own epic. This PR takes the cheaper and strictly-safe route instead: the SCIP Job fills the cache the warm path already reads, so a stale or missing index degrades to today's inline re-index rather than to anything new.

**Unverifiable in this environment:** the SCIP indexers were not run end to end here — no cluster, and rust-analyzer SCIP over the production workspace is a ~59-minute operation. `run_scip_index_command` mirrors `ensure_canonical_graph`'s indexer invocation argument-for-argument, and the renderer/binary argv contract proves the Pod's command line parses, but a first real run should be watched.

### DIFF
--- a/server/crates/djinn-agent-worker/src/main.rs
+++ b/server/crates/djinn-agent-worker/src/main.rs
@@ -215,6 +215,19 @@ enum Cmd {
         project_id: String,
     },
 
+    /// Run the SEMANTIC pass only (SCIP indexers) for a project and exit,
+    /// leaving its artifacts in the shared content-addressed SCIP cache.
+    ///
+    /// Dispatched by `djinn_k8s::scip_job::build_scip_index_job` in the
+    /// standalone SCIP-index Pod. Unlike `warm-graph` this runs no cargo warm
+    /// base, publishes no graph, and -- because it is not a
+    /// `LeaseIdentity::GraphWarm` consumer -- holds no build slot. The warm Job
+    /// picks its output up as a `CachedHit` and skips the 58m43s re-index.
+    ScipIndex {
+        /// Project id (matches `projects.id`). Positional.
+        project_id: String,
+    },
+
     /// Seed a private per-task-run Cargo target dir from the project's warm
     /// base and exit, printing a machine-readable summary line on stdout.
     ///
@@ -440,6 +453,7 @@ async fn run() -> Result<()> {
     match cli.cmd {
         Cmd::TaskRun(args) => run_task_run(args).await,
         Cmd::WarmGraph { project_id } => run_warm_graph(&project_id).await,
+        Cmd::ScipIndex { project_id } => run_scip_index(&project_id).await,
         Cmd::SeedCargoTarget { base, run_dir } => run_seed_cargo_target(&base, &run_dir),
         Cmd::CompareGraphArtifacts {
             project_id,
@@ -3655,6 +3669,83 @@ fn worker_bridge_ignores_pair(entity_type: &str, action: &str) -> bool {
 /// run `djinn_graph::canonical_graph::run_warm_graph_command` once and
 /// exit.  The heavy pipeline's progress lands in shared DB caches that
 /// the server process reads on subsequent graph queries.
+/// The `scip-index` subcommand: run the semantic pass and exit.
+///
+/// Deliberately a sibling of [`run_warm_graph`] rather than a flag on it. The
+/// two Pods have different capacity contracts -- the warm Job is charged a full
+/// build slot and this one is charged none -- so they must not be able to
+/// collapse into the same process by accident.
+async fn run_scip_index(project_id: &str) -> Result<()> {
+    #[cfg(target_os = "linux")]
+    {
+        return run_linux_warm_graph_with_initializer(
+            project_id,
+            initialize_linux_warm_lifecycle,
+            || run_scip_index_body(project_id),
+        )
+        .await;
+    }
+
+    #[cfg(not(target_os = "linux"))]
+    run_scip_index_body(project_id).await
+}
+
+/// The semantic-pass body. Mirrors [`run_warm_graph_body`]'s startup contract
+/// (volume readiness, DB bootstrap, `pre_anything` hooks) and then runs the
+/// indexers only.
+///
+/// The `pre_anything` hooks are NOT optional decoration here: a project pinning
+/// its toolchain relies on them for `rustup component add rust-analyzer`, and
+/// without that the SCIP indexers fail with "Unknown binary" -- which in this
+/// Pod would mean a completely empty index run rather than a degraded warm.
+/// They stay non-fatal for the same reason they are non-fatal on the warm path.
+async fn run_scip_index_body(project_id: &str) -> Result<()> {
+    let project_root = std::env::var("DJINN_PROJECT_ROOT")
+        .map(PathBuf::from)
+        .unwrap_or_else(|_| PathBuf::from(volume_contract::WORKSPACE_MOUNT_ROOT));
+    volume_contract::enforce_at_startup(
+        "scip-index",
+        &volume_contract::warm_roots(&project_root, project_id),
+    )?;
+
+    let db = bootstrap_warm_database().await?;
+    let ctx = WorkerWarmContext {
+        db,
+        indexer_lock: Arc::new(tokio::sync::Mutex::new(())),
+    };
+
+    let env_config_path = PathBuf::from(lifecycle::ENV_CONFIG_MOUNT_FILE);
+    match lifecycle::load_environment_config(&env_config_path).await {
+        Ok(Some(cfg)) => {
+            if let Err(e) =
+                lifecycle::run_phase(&project_root, "pre_anything", &cfg.lifecycle.pre_anything)
+                    .await
+            {
+                warn!(
+                    project_id,
+                    project_root = %project_root.display(),
+                    error = %format!("{e:#}"),
+                    "pre_anything hook failed; continuing with scip-index anyway"
+                );
+            }
+        }
+        Ok(None) => tracing::debug!(
+            project_id,
+            "no environment_config mounted at {} -- continuing without hooks",
+            env_config_path.display()
+        ),
+        Err(e) => warn!(
+            project_id,
+            error = %format!("{e:#}"),
+            "environment_config present but failed to load; ignoring"
+        ),
+    }
+
+    djinn_graph::canonical_graph::run_scip_index_command(&ctx, project_id)
+        .await
+        .with_context(|| format!("run_scip_index_command({project_id})"))
+}
+
 async fn run_warm_graph(project_id: &str) -> Result<()> {
     #[cfg(target_os = "linux")]
     {
--- a/server/crates/djinn-agent-worker/src/rendered_job_env_contract.rs
+++ b/server/crates/djinn-agent-worker/src/rendered_job_env_contract.rs
@@ -206,6 +206,65 @@ fn the_warm_pod_argv_parses_with_the_real_parser() {
     }
 }
 
+/// **The contract, SCIP-index path.** Same argument as the warm path, for the
+/// standalone semantic-index Pod added alongside it.
+///
+/// This Pod is dispatched by a scheduler on a 3-hour cadence rather than by an
+/// interactive trigger, so a `scip-index` subcommand that did not exist — or
+/// whose argv did not parse — would produce a Pod that exits 2, a Job that
+/// reports failure, and a change-detection ledger that never advances. That
+/// pattern would be invisible in the warm path's own tests, which is precisely
+/// why the derivation below reads the real clap parser rather than a list.
+#[test]
+fn the_scip_index_pod_argv_parses_with_the_real_parser() {
+    let job = djinn_k8s::scip_job::build_scip_index_job(
+        &KubernetesConfig::for_testing(),
+        "proj-scip",
+        "registry.example/djinn-project:scip",
+        "0123456789abcdef0123456789abcdef01234567",
+        None,
+    );
+    let pod = job
+        .spec
+        .as_ref()
+        .and_then(|spec| spec.template.spec.as_ref())
+        .expect("rendered SCIP Job has a pod spec");
+    let script = pod.containers[0]
+        .command
+        .as_deref()
+        .and_then(|command| command.last())
+        .expect("SCIP Pod runs a shell script");
+
+    let exec_line = script
+        .lines()
+        .find(|line| line.starts_with(&format!("exec {WARM_COMMAND_BIN}")))
+        .unwrap_or_else(|| panic!("SCIP Pod script never execs {WARM_COMMAND_BIN}:\n{script}"));
+
+    let argv = split_shell_words(exec_line.trim_start_matches("exec "));
+    assert_eq!(
+        argv.get(1).map(String::as_str),
+        Some("scip-index"),
+        "the SCIP Pod must invoke the semantic-only subcommand, not the combined \
+         warm pipeline — running `warm-graph` here would re-acquire the whole \
+         cargo phase this split exists to leave behind: {argv:?}"
+    );
+    Cli::try_parse_from(&argv).unwrap_or_else(|error| {
+        panic!(
+            "the SCIP Pod's own command line does not parse: {argv:?}\n{error}\nEvery SCIP \
+             Pod dispatched with this manifest exits 2 before indexing anything."
+        )
+    });
+
+    for arg in unsatisfiable_without("scip-index") {
+        assert!(
+            arg.is_positional(),
+            "required scip-index argument `{}` is not positional; the SCIP Pod's shell \
+             command supplies positionals only, so it must be rendered into the Pod env",
+            arg.get_id()
+        );
+    }
+}
+
 /// Minimal double-quote-aware tokenizer: enough for the argv this crate's own
 /// renderer emits (`exec <bin> warm-graph "<project id>"`), and it deliberately
 /// refuses anything more exotic rather than guessing at shell semantics.
--- a/server/crates/djinn-graph/src/canonical_graph.rs
+++ b/server/crates/djinn-graph/src/canonical_graph.rs
@@ -676,6 +676,124 @@ pub async fn run_warm_graph_command<C: WarmContext>(
     Ok(())
 }
 
+/// Run the **semantic pass only** and leave its artifacts in the shared
+/// content-addressed SCIP cache. Publishes no graph and writes no
+/// `repo_graph_*` row.
+///
+/// This is the body of the standalone SCIP-index Job
+/// (`djinn_k8s::scip_job`). The combined warm Job measured 90m42s end to end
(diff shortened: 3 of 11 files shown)


fernandobandeira:
Rebased onto current `main` (now includes #2695). Clean rebase, no conflicts — this PR does not touch `warm_job.rs` and touches `graph_warmer.rs` only for one accessor. `rendered_job_env_contract` now runs 5 tests (2695 added two, this PR added one); all pass together.

### One pre-existing failure, verified not from this PR

A full `cargo test -p djinn-agent-worker --bin djinn-agent-worker` run surfaced 6 failures. Triaged:

* **5 were load/parallelism flakes** — `lifecycle::tests::{nondjinn_multistep_db_preparation_fixture, pretask_activity_timeout_emits_blocked_and_environmental, pretask_enforces_timeout}`, `tests::warm_cargo_ordering_recorder_observes_prune_stamp_compile_and_end_sweep`, and `worker_services::lease_adapter_conformance_tests::direct_and_worker_rpc_run_the_same_lease_operation_script`. All pass when re-run serially.
* **1 is a genuine order-dependent test, and it fails on `main` too**: `tests::warm_shutdown_drains_two_registered_groups_and_setsid_adoptee`.

That last one passes 5/5 in isolation but fails whenever `tests::warm_cargo_ordering_recorder_observes_prune_stamp_compile_and_end_sweep` runs before it *in the same process* — including under `--test-threads 1`, so it is not a thread race. `PR_SET_CHILD_SUBREAPER` and process-group state are process-global, so the preceding test appears to leave reaping state that makes the drain assertion fail.

Reproduced on `origin/main` at 64e8952 (i.e. with #2695 merged and none of this PR applied), same panic and same surviving-pid shape:

```
thread ... panicked at crates/djinn-agent-worker/src/main.rs:6469:9:
all shutdown targets must drain: Err(WarmGraphShutdownError {
  surviving_pids: [2349905, 2349906], surviving_groups: [2349905, 2349906], source: None })
```

Flagging rather than fixing — it is unrelated to this change and belongs to whoever owns the warm shutdown/reaper tests.

PR #2726, opened 2026-07-28
test(agent): widen the JIT rollout env lock to every test that reads it

fernandobandeira:
Follow-up to #2725, which flagged this flake but deliberately left it out of that hotfix.

## The flake

`extension::tests::jit_trace_tests::*` fails intermittently. **Reproduced 1 run in 10 against a clean `origin/main` checkout** of the djinn-agent lib suite — this is pre-existing and not introduced by any recent change:

```
assertion `left == right` failed: unexpected JIT Prometheus delta for eligible_search
  left: 3
 right: 1
```

## Root cause

`JIT_PITFALLS_ENV_LOCK` was a file-private `Mutex` in `jit_trace_tests.rs`, scoped — per its own comment — so that *"concurrent JIT tests must not observe each other's env mutation"*. That scope was too narrow in two independent ways.

**1. The gate is read far outside the JIT tests.** `rollout_mode_from_env()` reads `DJINN_JIT_PITFALLS_ROLLOUT` at call time, and `maybe_append_pitfall_hint` runs on **three** handler paths — `call_write`, `call_edit`, and `call_apply_patch`. While a JIT test has the rollout `enabled`, every concurrently-running test on any of those paths also takes the JIT branch and bumps the shared process-global counter `djinn_jit_pitfall_hints_total{outcome="eligible_search"}`, corrupting the before/after delta in `assert_jit_pitfall_outcome_deltas`.

**2. Failures cascaded.** Because the guard was a `Mutex`, the first panic poisoned it and every subsequent JIT test panicked on `.lock().unwrap()`. One observed run reported 5 failures and another 3 — all from a single root failure, which badly obscures triage.

## Fix

Move the lock into `test_helpers` and make it an `RwLock`:

- JIT tests **mutate** the env → exclusive `write()`.
- The **68** write/edit/apply_patch tests across six files → shared `read()`, so they still run concurrently with each other and the added coverage costs no suite wall-clock.
- Both sides use `PoisonError::into_inner`, so a failure reports as itself instead of cascading.

No production code changes; the rollout gate itself is untouched.

## Verification

**25 consecutive full-suite runs clean** (`cargo test -p djinn-agent --lib`), against a baseline that fails ~1 in 10 — so ~2–3 failures would be expected over that many runs.

Worth recording: an earlier pass of this fix covered only `call_write` and **still failed on iteration 5**. That failure is what surfaced the `call_edit` / `call_apply_patch` paths. Stressing to 25 runs rather than stopping at the first few green runs is what made the difference.

`cargo clippy -p djinn-agent --lib --tests --no-deps` clean, `cargo fmt` clean.

### DIFF
--- a/server/crates/djinn-agent/src/extension/handlers/gate_guard/tests/edit_tests.rs
+++ b/server/crates/djinn-agent/src/extension/handlers/gate_guard/tests/edit_tests.rs
@@ -2,8 +2,10 @@ use super::*;
 
 // ─── AC 1: truncated read denies worker edit, leaves edit_forced unset ──
 
+#[allow(clippy::await_holding_lock)]
 #[tokio::test]
 async fn truncated_read_denies_worker_edit() {
+    let _jit_env = crate::test_helpers::jit_env_read_guard();
     let (worktree, state) = setup_worktree("gg-trunc-");
     let file = worktree.path().join("svc.rs");
     tokio::fs::write(&file, "fn main() {\n    println!(\"hello\");\n}\n")
@@ -65,8 +67,10 @@ async fn truncated_read_denies_worker_edit() {
 
 // ─── AC 1: uncovered read denies worker edit, leaves edit_forced unset ──
 
+#[allow(clippy::await_holding_lock)]
 #[tokio::test]
 async fn uncovered_read_denies_worker_edit() {
+    let _jit_env = crate::test_helpers::jit_env_read_guard();
     let (worktree, state) = setup_worktree("gg-uncovered-");
     let file = worktree.path().join("svc.rs");
     // File content: "AAAA" at bytes 0-4, "\n" at 4-5, "BBBB" at 5-9.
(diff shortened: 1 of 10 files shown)


fernandobandeira:
**Not ready — this does not fix the flake.** Recording the negative result rather than leaving the PR looking green.

Stress results against a baseline that fails ~1 run in 10:

| version | result |
|---|---|
| guards on `call_write` only | failed at iteration 5/20 |
| + `call_edit` / `call_apply_patch` (68 tests) | 0/25 — looked fixed |
| + `write_accepts_worker_role_plumbing` (the one inside `jit_trace_tests.rs`) | **4/25 — worse than baseline** |

So the `RwLock`-over-every-reader hypothesis is incomplete. The counter pollution has a source I haven't identified, and the middle run's 0/25 was luck rather than proof — which is exactly the trap of stopping at the first green stress run.

What is solid and worth keeping from this investigation:

- The failure is **pre-existing on `main`**, reproduced 1/10 on a clean checkout — it is not caused by #2725 or #2727.
- The mechanism is real: `rollout_mode_from_env()` reads a process-global env var at call time on the `call_write`/`call_edit`/`call_apply_patch` paths, and `assert_jit_pitfall_outcome_deltas` asserts deltas on a process-global Prometheus counter. Concurrent tests on those paths pollute it, and it is not limited to tests passing a `project_id` — `None` records `error`, and a disabled gate records `disabled_default_off`, both in the asserted set.
- Because the guard was a `Mutex`, one failure poisoned it and cascaded; observed runs failed 5 and 3 tests from a single root cause, which badly obscures triage.

I'd rather leave this open with the evidence than merge a change that adds 68 guards, splits a test file, and does not deliver the fix. The likely correct approach is to stop the pollution at the source — resolve the rollout mode per-`AgentContext` (defaulting to env in production, forced `Off` in the test helper) so non-JIT tests cannot enter the JIT branch at all. That costs a new field at 17 construction sites, which I judged too much blast radius to bundle into an outage hotfix, but it is the right shape for a standalone change.

PR #2768, opened 2026-07-29
ci: run the orphaned deploy contract suites

fernandobandeira:
## Five deploy contract suites have never executed in CI — not once

`scripts/test-helm-chart-contracts.sh` globs exactly one directory. Its own header promises that the directory is "globbed, never enumerated: a new contract script dropped in [runs automatically]" — and that promise is real, but it stops at `deploy/helm/djinn/tests/`. Every contract suite written anywhere else in `deploy/` is unreachable from any workflow.

```
$ grep -rn "runbooks/tests\|node/k3s/tests\|preflight/tests" .github/workflows/ scripts/
(no output)
```

Orphaned, on `main`, today:

| Suite | Status |
| --- | --- |
| `deploy/node/k3s/tests/containerd-config-version-detection.sh` | never run in CI |
| `deploy/node/k3s/tests/conformance-version-lifecycle.sh` | never run in CI |
| `deploy/runbooks/tests/cgroup-launcher-rearm.sh` | never run in CI |
| `deploy/runbooks/tests/bzpt-production-activation.sh` | **never run in CI — and it was one of proposal d308's acceptance criteria.** It has been dead since d308 landed. The AC was attested against a suite that has executed zero times. |
| `deploy/preflight/tests/render-gate.sh` | never run in CI |

That last row is the whole point. A merged suite that no lane invokes is not coverage; it is the appearance of coverage. All five pass today, so nothing is broken right now — which is exactly the window in which to wire them up, before one of them silently starts failing and nobody finds out.

## What this adds

**`scripts/test-deploy-contracts.sh`** (new) — modelled directly on `scripts/test-helm-chart-contracts.sh`: same per-script `PASS`/`FAIL` roster, same `::group::` framing, same ungrouped output on failure so a red CI log is diagnosable without expanding anything, same non-zero exit on any failure.

Two properties it does not inherit but needs:

- **It globs directories, never files.** `deploy/node/k3s/tests/*.sh` and `deploy/runbooks/tests/*.sh` by default, non-recursive (matching the existing runner's deliberate choice). A new suite dropped into either directory runs with no edit to the runner or the workflow.
- **An empty roster is a hard failure.** A silently empty glob is precisely the failure mode this PR exists to fix; a runner that reports success because it found nothing is worse than no runner, because it manufactures the evidence that the suites ran. Both a missing directory and a present-but-empty one exit 1.

It also takes optional directory arguments, so the preflight directory can be hosted on a different lane while still being **globbed rather than enumerated** — the guarantee holds for all three directories, not just the two that share a lane.

## Lane placement

**`local-dev-contracts`** hosts the four shell-only suites. They need nothing but bash and the checkout. The runner is also added to that lane's existing `bash -n` syntax-check list.

**`server-clippy`** hosts `deploy/preflight/tests/render-gate.sh`, because that suite is not shell-only: it runs `cargo build -p djinn-k8s --bin render-gate` and shells out to `helm template`. `local-dev-contracts` has helm but no Rust, and putting it there would mean installing a cold toolchain and doing a cold `djinn-k8s` build on a lane whose whole design is "no toolchain". `server-clippy` already runs `rustup toolchain install` and restores the `server-test` `Swatinem/rust-cache` family — which, per that lane's own comment, now carries *both* the check-mode and build-mode dependency sets — so the build is incremental there. It gains one `azure/setup-helm@v4` step, pinned to **v3.12.3**, the same version `local-dev-contracts` installs: two lanes rendering the same chart on different helm majors would be a contract that means nothing.

The step runs last in the job so a deploy-contract failure is never confused with a clippy failure and cannot delay the lint signal. Headroom is fine: recent `Server Clippy` runs finish in ~4 minutes against a 25-minute timeout, and the local `render-gate` build measured 1m06s warm.

No routing change is needed. `scripts/ci-changed-scope.mjs` is fail-closed — paths under `deploy/node/` and `deploy/runbooks/` are unclassified, which selects full validation, which selects both lanes.

## Verification

All four shell suites pass through the runner (34 + 31 + 1 + 26 assertions), and `deploy/preflight/tests/render-gate.sh` passes through it as well (12 cases, including the stock-chart-defaults case).

The runner was proven non-vacuous two ways, both restored afterwards:

- **Empty roster fails.** Emptying `deploy/runbooks/tests/` and running the default invocation: `FAIL: no deploy contract scripts found in .../deploy/runbooks/tests`, exit 1. It does not report "2 scripts passed".
- **A broken assertion fails.** Mutating one `expect_eq` in `containerd-config-version-detection.sh` produced exit 1, a `FAIL(exit 1)` roster row naming that exact script, and `FAIL: 1 of 4 deploy contract scripts failed`.

Also: `bash -n` clean on the new script, `actionlint -shellcheck= -pyflakes=` clean on the workflow, and the workflow YAML parses.

Diff is two files: the new runner and `quality-gate.yml`. No suite, no file under `deploy/`, and not `test-helm-chart-contracts.sh` was touched.

> Vercel preview checks are currently rate-limited and will show red. They are not required checks.

### DIFF
--- a/.github/workflows/quality-gate.yml
+++ b/.github/workflows/quality-gate.yml
@@ -273,13 +273,24 @@ jobs:
             scripts/test-tiltfile-config.sh \
             scripts/test-wrap-agent-runtime-image.sh \
             scripts/test-helm-github-app-render.sh \
-            scripts/test-helm-chart-contracts.sh
+            scripts/test-helm-chart-contracts.sh \
+            scripts/test-deploy-contracts.sh
       - name: Test content-addressed warm builds
         run: bash scripts/test-tilt-warm-restart.sh
       - name: Test isolated Tilt configuration
         run: KUBECONFIG=/dev/null sh scripts/test-tiltfile-config.sh
       - name: Test runtime-base self-healing wrapper
         run: bash scripts/test-wrap-agent-runtime-image.sh
+      # deploy/node/k3s/tests/ and deploy/runbooks/tests/ were unreachable from
+      # any workflow until this step existed: scripts/test-helm-chart-contracts.sh
+      # globs only the chart's own tests directory, so four suites had never run
+      # in CI once — including deploy/runbooks/tests/bzpt-production-activation.sh,
+      # which was an acceptance criterion of proposal d308. These need nothing
+      # but bash, which is why they are hosted here rather than on a Rust lane.
+      # The runner globs directories and fails on an empty roster, so a new
+      # suite dropped into either directory runs here with no edit to this file.
+      - name: Test deploy contracts (node, runbooks)
+        run: bash scripts/test-deploy-contracts.sh
       - name: Lint local and default Helm values
         run: |
           helm lint deploy/helm/djinn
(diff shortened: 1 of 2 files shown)


fernandobandeira:
### CI evidence from the first run (30479829505) — both new steps green

| Step | Lane | Result |
| --- | --- | --- |
| `Test deploy contracts (node, runbooks)` | Local Dev Contracts | **success** |
| `Test deploy render-gate contract` | Server Clippy | **success** |

`Local Dev Contracts`, `Server Clippy`, and `CI Contracts` (which is what runs `actionlint` over this workflow) all passed. `Preflight routing` selected full validation, as expected: `scripts/ci-changed-scope.mjs` is fail-closed and `deploy/` paths are unclassified.

### Follow-up commit: `Server Clippy` timeout 25 -> 35

Hosting the render-gate contract here is not free, and the first run measured the real cost:

```
Clippy                        3m21s   (18:26:29 -> 18:29:50)
Test deploy render-gate       7m42s   (18:29:50 -> 18:37:32)
job total                    11m50s   (was ~4m9s before this PR)
```

7m42s is well above the ~1m06s the same suite takes locally, and the reason is the one thing this lane's own comments are emphatic about: `cargo clippy` emits **check-mode** artifacts for workspace crates, so `cargo build -p djinn-k8s --bin render-gate` has to compile the whole `djinn-db -> djinn-core -> djinn-k8s` chain in **build mode**. Only the third-party dependencies come from the restored cache.

That still makes this the cheapest available host — anywhere without the restored cache would also pay for the ~4700 third-party units — but it leaves 11m50s inside a 25-minute timeout, which is thinner than this lane was sized for and would not survive a cold-cache run. Raised to 35 with the measurement recorded inline.

PR #2799, opened 2026-07-30
ci: cut PR-lane wall clock and stop the cache budget bleeding

fernandobandeira:
## Why

The PR lane and the merge-queue lane run an **identical job set** — every guard in `quality-gate.yml` is `github.event_name != 'push'` — and both sit at a p50 of **14.1 min** (n=65 PR / n=98 merge_group successful runs). The queue is authoritative, so the PR lane's only job is to predict churn cheaply. Its latency is what's worth attacking.

Measured critical path, run [30498632438](https://github.com/djinnos/djinn/actions/runs/30498632438) (typical PR):

| Segment | Time | % |
|---|---|---|
| queue + `preflight` | 0:27 | 3% |
| shard setup (container, checkout, cache restore 38s, migrations 35s) | 1:36 | 11% |
| **`Build test binaries`** (`--workspace --all-targets`) | **5:54** | **42%** |
| **`Run planner-selected shard tests`** (2909 of 11 633) | **4:00** | **28%** |
| upload + `Quality Gate` rollup | 1:10 | 8% |

### 1. Split the deploy render-gate contract off `Server Clippy`

`Server Clippy` was 11m50s — the second-longest lane — of which **only 3m21s was clippy**. The other 7m42s was `Test deploy render-gate contract`, a Helm/shell suite running as a serial tail. Two independent gates were queued behind one runner: clippy's lint signal was delayed by a Helm contract, and the contract couldn't start until the lint finished.

Now `deploy-render-gate`, routed by the same `preflight.outputs.clippy` output and wired **fail-closed** into the `quality-gate` rollup (`needs` + a `check` line). Clippy returns to ~4m; the contract runs in parallel at ~8.5m, inside the server-test critical path.

The split does **not** make the 7m42s cheaper, and the old comment misattributed it. It blamed `cargo clippy` emitting check-mode artifacts, but check-mode `.rmeta` and build-mode `.rlib` units coexist under different `-C metadata` and neither displaces the other. The real cause is that `actions/checkout` restamps every workspace source mtime, so cargo rebuilds the `djinn-db → djinn-core → djinn-k8s` chain in build mode on *any* lane. What the split removes is paying it in series.

### 2. `PR_MAX_SHARDS` 4 → 8

Only the test-**run** half of a shard scales with shard count; the 5m54s build is the same `--workspace` compile in every shard and doesn't shrink when the shard's test list does. Doubling shards halves ~4m and leaves ~7m30s untouched. Stops at 8 because the fixed half then dominates — 16 would buy progressively less for a full extra runner each time.

Merge-queue stays at `WIDE_DEFAULT_SHARDS = 4` deliberately: a failing queue entry ejects and invalidates the entries speculatively stacked behind it (`maximumEntriesToBuild=5`, `mergingStrategy=HEADGREEN`), so the queue's cost function is throughput under contention, not latency of one entry.

The exact-once proof is shard-count agnostic, so this is a balancing knob, not a coverage one. 48/48 planner tests pass.

### 3. Drop `type=gha` buildkit cache from `release.yml`

It could never work. GitHub scopes a cache entry to the ref that wrote it, and a run may read only its own ref or the default branch — and this workflow triggers on **tags**. Verified: each tag's 44 blob entries were created *and* last-read entirely inside that tag's own run.

| Tag | Release run | Blobs created | Last read |
|---|---|---|---|
| v0.7.25 | 15:45 → 16:11 | 15:45–15:48 | **15:50** |
| v0.7.26 | 18:20 → 18:46 | 18:21–18:22 | **18:25** |
| v0.7.27 | 19:36 → 20:02 | 19:37–19:39 | **19:41** |

v0.7.26's run never touched v0.7.25's blobs. That was 0.92 GB written per release, ~3 releases/day, **evicting the `server-test` family that every CI lane restores** — the eviction is the reason to remove it, not the wasted bytes.

Nothing is lost. `runtime-base` builds the same image twice, and the second build reuses the first's layers from the buildx **builder's** local cache (it persists for the job's lifetime); intra-run reuse never depended on an exported cache. Single-build jobs got nothing from it at all.

Real cross-release reuse needs a ref-independent backend (`type=registry,ref=…:buildcache` on GHCR, which this workflow already authenticates to). That's a release-**speed** change with its own export cost to measure on a production path — deliberately not bundled here.

### 4. `cache-prune`: fix the budget report, `--keep` 2 → 1

The report step had failed on **every scheduled run since at least 2026-07-28**:

```
the `--slurp` option is not supported with `--jq` or `--template`
```

A newer `gh` on the runner image began rejecting that combination. The prune itself was unaffected and kept working (run 30429331029 reclaimed 2.46 GB before dying here), so the only symptoms were a red workflow and — worse — the budget notice this step exists to emit never reaching anyone while the repo sat at **10.49 GB against a 10 GB cap**. Rewritten onto the unpaginated `/actions/cache/usage` endpoint, so there is no `--paginate` and no `--slurp` left for the incompatibility to apply to.

`--keep 2` protected nothing: rust-cache restores by key **prefix** and GitHub returns the newest match, so the older entry is never read — including in the case it looks like insurance against, a corrupt newest entry (`cache-on-failure: true` means a partial deposit is saved and still wins the prefix match). The script's own header already says only the newest entry per family is ever read. That was 1.67 GB + 0.78 GB = **2.45 GB of pure ballast**. Safe because each family has exactly one writer, on main, in a push/dispatch-only job — enforced by `ci-cache-policy.test.mjs`.

Note both defaults had to move: `inputs.keep` is empty on the scheduled run, so the `|| '1'` fallback, not `default:`, governs the daily prune.

## How to measure

Baseline (before this PR), from the last 100 completed runs of each event:

| Metric | Baseline |
|---|---|
| PR-lane wall clock | **p50 14.1 / p75 14.4 / p90 16.1 min** (n=65) |
| merge_group wall clock | p50 14.1 / p75 15.1 / p90 17.2 min (n=98) |
| `Server Clippy` | 11m50s (clippy 3m21s + render-gate 7m42s) |
| `Server Test` shard | 12–13m (setup 1m36s + build 5m54s + run 4m00s) |
| Actions cache | 10.49 GB / 154 entries, over a 10 GB cap, evicting |

Reproduce after merge:

```bash
# per-run wall clock distribution
gh api "repos/djinnos/djinn/actions/workflows/quality-gate.yml/runs?event=pull_request&per_page=100&status=completed" \
  -q '.workflow_runs[] | select(.conclusion=="success") | "\(.run_started_at) \(.updated_at)"'

# per-job and per-step timings for one run
gh api repos/djinnos/djinn/actions/runs/<id>/jobs \
  -q '.jobs[] | "\(.name)\t\((((.completedAt|fromdate)-(.startedAt|fromdate))/60)|floor)m"'

# cache budget
gh api repos/djinnos/djinn/actions/cache/usage
```

Expected: PR-lane p50 ≈ **11.3 min** (−20%), `Server Clippy` ≈ 4m, cache settling ≈ 5.3 GB once the prune runs and no new tag-scoped blobs are written.

## What was measured and rejected

- **Crate-scoped test selection.** Reverse-dependency closure over the last 48 server-touching commits retains a **median 70%** of test time, ≥90% for 22 of 48, and **never below 50%**. `djinn-db` is in ~40% of PRs and pulls in 18 of 32 packages. Best case ~1.2 min off 14, against forfeited signal — of 100 failing PR runs, 66 failed on a test shard and **45 were caught by nothing else**.
- **Crate-scoped clippy.** Only 3m21s to begin with, and narrowing the package set changes the resolver-2 feature union, hands shared deps a different `-C metadata`, and misses the ~1900 cached units — the trap already documented at `quality-gate.yml:768-784` and measured in #2349.
- **`cache-workspace-crates`.** Measured in #2349: cache 1267 → 2283 MB, shard build time unmoved.
- **sccache.** The config comment's objection (forbids incremental) is scoped to the platform pods; CI already pins `CARGO_INCREMENTAL: 0`. But it's still the wrong tool: it does not cache **linking**, and the workspace ships **175 test binaries** against 41 crate compiles. A `cargo build --timings` run is needed to get the compile/link split before investing.
- **mold / lld.** mold needs a network install that has broken CI before. `rust-lld` ships with the toolchain, but the stable flag is unavailable on the pinned channel (`-C link-self-contained=+linker` requires `-Z unstable-options` on 1.97.1).

## Verification

- `actionlint -shellcheck= -pyflakes=` clean on all workflows
- All 9 locally-runnable `ci-contracts` suites pass, including `ci-cache-policy` (single-writer ownership for the new lane) and `ci-release-image-validation` (release step ordering preserved)
- `ci-nextest-plan.test.mjs` 48/48
- Rollup verified structurally: `deploy-render-gate` in `needs`, in `env`, and in a `check` line

### DIFF
--- a/.github/workflows/cache-prune.yml
+++ b/.github/workflows/cache-prune.yml
@@ -26,7 +26,23 @@ on:
       keep:
         description: Newest entries to retain per rust cache family
         type: string
-        default: '2'
+        # 2 -> 1. Keeping a second entry per family protected nothing: rust-cache
+        # restores by key PREFIX and GitHub returns the newest match, so the
+        # older entry is never read under any circumstance — including the one it
+        # looks like insurance against, a corrupt newest entry (the warm job sets
+        # `cache-on-failure: true`, so a partial deposit is saved and still wins
+        # the prefix match). This script's own header already says only the
+        # newest entry per family is ever read.
+        #
+        # It was not free: measured 2026-07-30, the second copies were 1.67 GB
+        # (server-test) and 0.78 GB (server-aarch64-check) = 2.45 GB of a 10 GB
+        # repo cap that was sitting at 10.49 GB and actively evicting the live
+        # entries CI restores on every run.
+        #
+        # Safe because each family has exactly one writer, on main, in a
+        # push/dispatch-only job (`ci-cache-policy.test.mjs` enforces this), and
+        # an in-flight save is not visible to restores until it completes.
+        default: '1'
 
 # The token needs `actions: write` to delete a cache. Everything else stays
 # read-only: this workflow must never be able to touch repository contents.
@@ -56,17 +72,35 @@ jobs:
           GH_TOKEN: ${{ github.token }}
           GITHUB_REPOSITORY: ${{ github.repository }}
         run: |-
+          # Both defaults must agree. `inputs.keep` is empty on the SCHEDULED
+          # run (inputs exist only for workflow_dispatch), so this fallback —
+          # not the `default:` above — is what governs the daily prune. They
+          # were 2 and 2; see the input's comment for why 1.
           node scripts/prune-actions-caches.mjs \
-            --keep '${{ inputs.keep || '2' }}' \
+            --keep '${{ inputs.keep || '1' }}' \
             ${{ inputs.dry_run && '--dry-run' || '' }}
+      # This step failed on every scheduled run from at least 2026-07-28 until
+      # 2026-07-30 with:
+      #
+      #   the `--slurp` option is not supported with `--jq` or `--template`
+      #
+      # A newer `gh` on the runner image began rejecting that flag combination.
+      # The prune itself was unaffected and kept working (run 30429331029
+      # reclaimed 2.46 GB before dying here), so the only symptom was a red
+      # workflow and — worse — the budget notice this step exists to emit never
+      # reaching anyone while the repo sat over its cap.
+      #
+      # Rewritten onto the dedicated cache-usage endpoint rather than re-fixing
+      # the flags: it returns both figures in ONE unpaginated response, so there
+      # is no `--paginate`, no `--slurp`, and nothing left for that
+      # incompatibility to apply to.
       - name: Report remaining budget
         if: always()
         env:
           GH_TOKEN: ${{ github.token }}
         run: |-
-          total=$(gh api "/repos/${{ github.repository }}/actions/caches?per_page=100" --paginate --slurp \
-            --jq '[.[].actions_caches[].size_in_bytes] | add // 0')
-          count=$(gh api "/repos/${{ github.repository }}/actions/caches?per_page=100" --paginate --slurp \
-            --jq '[.[].actions_caches[]] | length')
+          usage=$(gh api "/repos/${{ github.repository }}/actions/cache/usage")
+          total=$(printf '%s' "$usage" | jq -r '.active_caches_size_in_bytes // 0')
+          count=$(printf '%s' "$usage" | jq -r '.active_caches_count // 0')
           gb=$(awk "BEGIN { printf \"%.2f\", ${total}/1073741824 }")
           echo "::notice::cache-budget ${gb} GB across ${count} entries (repo cap 10 GB, LRU eviction)"
--- a/.github/workflows/quality-gate.yml
+++ b/.github/workflows/quality-gate.yml
@@ -633,14 +633,13 @@ jobs:
     needs: preflight
     if: needs.preflight.outputs.clippy == 'true' && github.event_name != 'push'
     runs-on: ubuntu-latest
-    # Raised from 25 when the deploy render-gate contract moved onto this lane.
-    # Measured on run 30479829505: Clippy 3m21s, render-gate 7m42s, job 11m50s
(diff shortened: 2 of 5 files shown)


fernandobandeira:
Heads up for whoever reviews the shard-count change: **\`PR_MAX_SHARDS = 8\` in this PR is inert on its own.**

This PR's own CI run ([30538880312](https://github.com/djinnos/djinn/actions/runs/30538880312)) still ran 4 shards. The planner cold-starts on every run — its timing-artifact restore has never once succeeded — and `pickShardCount` short-circuits on `timings.size === 0` before it looks at the profile, so `PR_MAX_SHARDS` is unreachable code today.

Two stacked bugs cause it (an artifact filter on a field that endpoint does not return, and a `fromdateiso8601` call that rejects the millisecond precision the writer emits). Both are fixed in #2803, which is independent and composable in either merge order:

- this PR alone → 4 shards, `PR_MAX_SHARDS=8` sits unused
- #2803 alone → 4 shards but duration-balanced instead of count-balanced
- both → 8 shards, ~1456 tests / ~151s each

Everything else in this PR is verified working from that same run: `Server Clippy` **11m50s → 4m0s**, and the split-out `Deploy Render Gate` ran in parallel at 8.7m, inside the server-test critical path.

PR #2804, opened 2026-07-30
ci: add a dispatch-only cargo --timings measurement lane

fernandobandeira:
## Why

Every remaining build-time decision is blocked on one unmeasured number. Each `server-test` shard spends **~5m54s** in `Build test binaries` — now ~62% of the shard, and the largest single item on the PR critical path. How much of it is **compiling** and how much is **linking** decides between two mutually exclusive investments:

- **library / compile-dominated** → a content-addressed compile cache (sccache on an external backend — the Actions cache is capped at 10 GB with no room) is worth the work.
- **binary / link-dominated** → sccache is near-worthless, since it does not cache linking. The lever is instead the binary **count**: nextest lists **175 test binaries** against only **41** workspace crate compiles, because rustc builds one per file in `tests/` (`djinn-db` ships 40, `djinn-control-plane` 32).

Picking wrong wastes the expensive one. This lane measures it instead.

Deriving it from a job log does not work, and was tried: cargo prints `Compiling <pkg>` at unit **start**, so links of earlier binaries overlap later compiles, and the anchors straddle two cargo invocations inside one step — yielding a 395s "span" for a 358s step and a *negative* link tail.

## Shape

An **instrument, not a gate**:

- `workflow_dispatch` only, via a new `cargo_timings` boolean input, default **false** — zero cost on every normal PR, merge-queue and main run.
- Deliberately **absent from the `quality-gate` rollup**, so it asserts nothing and can never block a merge.
- Restores `server-test` **read-only** (`save-if: false`), so it measures what a shard actually rebuilds from a warm cache. Measuring a cold build would answer a question nobody is asking.
- Carries the standard runner-disk preflight — it builds the same ~11–12 GiB of test binaries a shard does.
- Uploads the `--timings` HTML as an artifact, and prints a summary so the answer is visible without downloading anything.

## The summariser is validated, and that mattered

I ran it against a real cargo 1.97.1 report before shipping. The obvious reading of the format is wrong in three ways:

| Assumption | Reality |
|---|---|
| `rmeta_time` gives the type-check vs codegen split | **The field does not exist.** An earlier draft aggregated it and would have printed `100% codegen+link` for every possible input |
| `mode` groups units usefully | It is the literal string `"todo"` |
| `sections` gives frontend/codegen per unit | Only for plain **library** units. Every link-producing unit — all 175 test binaries, every bin — returns `sections: []` |

That first one is the reason this check was worth doing: it would have shipped an *answer-shaped non-answer*, which is worse than no measurement, because it looks like data.

**So `--timings` cannot isolate link time, and this step no longer claims to.** It reports per-unit durations grouped into:

- **link-producing** units — test binaries and bins, where compile and link are fused and sccache cannot help the link half
- **library** units — rlib/rmeta, which a content-addressed cache can serve outright

plus the frontend/codegen split wherever cargo does report it, and the 20 slowest units.

Read it as: library-dominated → sccache is worth pursuing. Binary-dominated → the 175-binary count is the lever, not caching.

Verified end-to-end on a synthetic crate with a bin, a lib and three integration tests:

```
units=9  sum-of-unit-durations=1.9s  (wall clock is lower: units run in parallel)
  test binary (compile+LINK)   n=   5       1.3s   67.7%
  build-script / dep lib       n=   3       0.4s   21.9%
  bin (compile+LINK)           n=   1       0.2s   10.4%

frontend/codegen split, where cargo reports it (3/9 units): codegen=0.1s, frontend=0.3s
```

## How to run it

```bash
gh workflow run quality-gate.yml --ref main -f cargo_timings=true
```

Then read the `cargo-build-timings` artifact, or the job's own summary output.

## Note on `nextest`

Uses `cargo build --workspace --all-targets --features qdrant --timings` rather than `cargo nextest run --no-run`, because nextest has no `--timings` passthrough. It builds the same test binaries, so the split it reports is the one the shards pay.

## Verification

- `actionlint -shellcheck= -pyflakes=` clean
- All 9 locally-runnable `ci-contracts` suites pass, including `ci-cache-policy` (the new lane is a restore-only consumer of a declared family) and the `workflow_dispatch` reachability assertion, which still holds now that the trigger carries inputs
- Heredoc verified to survive YAML block-scalar stripping (terminator lands at column 0); extracted shell passes `bash -n` and the embedded Python passes `py_compile`
- Confirmed absent from the rollup `needs`, and skipped on `pull_request`

### DIFF
--- a/.github/workflows/quality-gate.yml
+++ b/.github/workflows/quality-gate.yml
@@ -7,7 +7,15 @@ on:
     branches:
       - main
   merge_group: null
-  workflow_dispatch: null
+  workflow_dispatch:
+    inputs:
+      cargo_timings:
+        description: >-
+          Also run the Cargo Build Timings measurement lane (see that job).
+          Off by default: it is an instrument, not a gate, and it pays a full
+          workspace build.
+        type: boolean
+        default: false
 concurrency:
   group: ${{ github.event_name == 'pull_request' && format('ci-pr-{0}', github.event.pull_request.number) || github.event_name == 'push' && 'ci-main' || github.event_name == 'merge_group' && format('ci-merge-group-{0}', github.event.merge_group.head_sha || github.ref) || format('ci-dispatch-{0}', github.run_id) }}
   cancel-in-progress: ${{ github.event_name != 'merge_group' }}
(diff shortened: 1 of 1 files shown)


fernandobandeira:
Closing — the question this lane existed to answer has been answered, by running `cargo build --timings` locally instead.

**Result: the build is link-bound.** Filtered to the 225 workspace units CI rebuilds every job (1367s of unit-time):

| unit kind | n | time | share |
|---|---|---|---|
| test binary (compile+LINK) | 175 | 1042s | **76.2%** |
| dep lib / build-script | 31 | 271s | 19.8% |
| bin (compile+LINK) | 10 | 39s | 2.8% |
| library (rlib/rmeta) | 3 | 13s | 1.0% |

**79% link-producing, 21% library/cacheable.** So sccache is the wrong investment: it tops out at that 21%, and cannot help the link half of any test binary. The lever is the 175-binary count — `djinn_db "lib" (test)` alone is the largest single workspace unit at 75s.

Caveats: cold build on 16 cores with another `cargo test` running concurrently, so absolute seconds are inflated; the ratio is the usable output. "compile+LINK" is fused, so 76.2% is not purely link — the point is that it is the category where sccache structurally cannot help.

Worth preserving from this branch, since the lane is going away — `cargo --timings` output has three traps, all verified against a real cargo 1.97.1 report:

1. There is **no `rmeta_time` field**. Aggregating it (my first draft) yields "100% codegen+link" for every possible input — an answer-shaped non-answer.
2. `mode` is the literal string `"todo"`; grouping by it says nothing.
3. `sections` carries a frontend/codegen split, but **only for plain library units** — every link-producing unit returns `sections: []`. So `--timings` cannot isolate link time at all.

If a warm-case, reproducible-in-CI version is ever wanted, this branch is the starting point.

PR #2806, opened 2026-07-30
fix(ci): provision 8 server-test shards so half the suite stops being skipped

fernandobandeira:
## Severity

**Since #2803 merged (12:23Z), every PR-lane run has executed half its tests and reported green.**

`server-test`'s matrix is static — the shards deliberately do not `needs: nextest-plan`, so they can build concurrently with the planner's discovery — and it had **four** rows. #2799 raised `PR_MAX_SHARDS` 4 → 8 without widening it. That was inert until #2803 repaired the timing restore and let the planner actually reach `PR_MAX_SHARDS`. From then on:

```
run 30544309199 (pull_request, 12:51Z)
  plan shardCount=8, 11666 tests, exactOnce=true
  jobs created:  shardIndex 0,1,2,3
  shardIndex 4-7: 5834 of 11666 tests (50%)  ->  NEVER RAN
  run conclusion: success
```

## Why it was silent

Every safeguard was pointed the other way:

- The **exact-once proof** validates the *plan's* partition, not what executed — so it passed.
- The four provisioned shards each resolved their row and passed normally.
- **No job can detect its own absence.** The shards that would have covered indices 4–7 are exactly the ones that were never created.
- The existing surplus-index guard handles plan **narrower** than matrix. The deficit direction had no check.

The workflow even documented the invariant it depended on — *"the matrix is statically provisioned at the planner's maximum width … PR\_MAX\_SHARDS … are all 4"* — and that comment silently stopped being true when the constant changed.

## The fix

Widening the matrix alone would leave the same trap armed for whoever next edits a shard constant, so:

1. **Matrix → 8 rows** (shardIndex 0–7).
2. **Fail-closed runtime guard.** Every shard asserts `plan.shardCount <= CI_SERVER_TEST_MATRIX_WIDTH` and fails loudly if the plan is wider. Checked in the shards that *do* exist, precisely because the missing ones cannot check anything.
3. **Contract test.** Asserts matrix width ≥ `max(PR_MAX_SHARDS, WIDE_DEFAULT_SHARDS, COLD_START_SHARDS)`, that shardIndex values form a dense 0-based range, that `CI_SERVER_TEST_MATRIX_WIDTH` equals the real row count, and that the runtime guard is present.

## Verified non-vacuous

Reverting the matrix to four rows makes the new test fail:

```
✖ provisions a server-test matrix at least as wide as the widest possible plan
  AssertionError: server-test's static matrix has 4 rows but the planner can emit
  up to 8 shards (max of PR_MAX_SHARDS=8, WIDE_DEFAULT_SHARDS=4, COLD_START_SHARDS=4).
  Tests on the unprovisioned tail indices would never run and CI would still pass.
```

Restoring it: 50/50 pass. `actionlint` clean; all other contract suites pass.

## Affected runs

Only `pull_request` runs started after 12:23:16Z: 30544309199 and 30544818057 (#2805). Both need a rerun after this lands — their green is not trustworthy. Merge-queue runs were never affected (`WIDE_DEFAULT_SHARDS` stayed at 4).

### DIFF
--- a/.github/workflows/quality-gate.yml
+++ b/.github/workflows/quality-gate.yml
@@ -52,6 +52,12 @@ env:
   # what a healthy hosted runner has free after the reclaim set below, so a
   # healthy runner never trips it — a preflight that false-fails is worse than
   # no preflight.
+  # Number of rows in `server-test`'s static matrix. Asserted against the plan
+  # inside every shard, so a plan wider than the matrix fails loudly instead of
+  # silently dropping the tail shards' tests. Must equal the row count of that
+  # matrix AND be >= the planner's widest shard count; `ci-nextest-plan.test.mjs`
+  # enforces both.
+  CI_SERVER_TEST_MATRIX_WIDTH: "8"
   CI_RUNNER_DISK_MIN_GB: "14"
   # Warn band. Free space between MIN and WARN still completes the build but is
   # already abnormal for a hosted runner, so annotate it: the run summary is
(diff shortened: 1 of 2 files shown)


fernandobandeira:
**Correction to this PR's description — the failure mode was not silent, and the blast radius was one run.**

I claimed the coverage gap produced a green run. It did not. Verified after merge:

- Run 30544309199 concluded **`failure`**, not success. `Publish Nextest Timing` failed its "Merge current successful shard timings" step, which iterates `for (let i = 0; i < proof.shardCount; i++)` and so demands a timing artifact for all 8 planned shards. Only 4 existed, the assertion tripped, and `Quality Gate` failed with it. That is precisely the safeguard the `server-test` comment promised ("the timing-publish job still verifies exact-once coverage across the shards that ran") — it worked.
- **Only one run was ever affected**, and it was this branch's own predecessor (`ci-cargo-build-timings`). A run needed BOTH #2803 (so the planner reaches PR_MAX_SHARDS=8) AND a 4-row matrix — a 41-minute window between 12:23:16Z and 13:04:20Z.
- **#2805 was never affected.** Its head predates #2803, so its planner cold-starts to `COLD_START_SHARDS=4`, matching its 4-row matrix.

So the accurate framing: 50% of tests genuinely did not run in that one run, but CI **blocked** it rather than passing it. What this PR fixes is real — the matrix must be as wide as the widest plan — and the new guard still earns its place, because it converts an obscure downstream assertion in a different job into a named error at the point of cause. But "silently halved the test suite and reported green", in the commit message and the description above, is wrong.

PR #2824, opened 2026-07-30
test: give metrics-asserting tests a recorder of their own

fernandobandeira:
Two pre-existing flake clusters that fail at default test concurrency, one root cause each. Test-only, plus one new test-support helper. No production behaviour is touched.

## Root cause

The `metrics` recorder installed by `djinn_telemetry::init()` is process-global, and cargo runs every test in a binary as a thread of one process. A test that reads that registry is reading whatever the **whole binary** emitted, so it is coupled to every concurrently running sibling.

**`djinn-supervisor`** — the nine workspace clone/cleanup telemetry tests took a before/after delta around the operation under test. A delta narrows the window but does not close it: any sibling that clones or tears down a workspace lands inside it. That is why the counts were over by exactly the concurrent-sibling count (`7.0 vs 4.0`, `17.0 vs 14.0`, `18.0 vs 15.0`). The two `*_TELEMETRY_MUTEX`es only excluded the tests holding them, so they bought serialization without determinism.

**`djinn-coordinator`** — `assert_strike_decision_labels_are_bounded` read the whole registry and asserted its bounded-label allowlist against every series the binary had ever emitted. A sibling exercising `environmental_restart_orphan` (a legitimate, equally bounded source that these two call paths never emit) failed *these* tests instead of its own.

## Change

New `djinn_telemetry::IsolatedRecorder` — the `async` counterpart of the existing `render_isolated`, which only works for synchronous closures. Construct it, hold the guard returned by `scope()` across the region to capture, then `render()` a registry no other test can reach. Assertions become absolute rather than deltas, both mutexes are deleted, and the whole group runs in parallel.

Soundness rule is documented on the type and is the same as `render_isolated`'s: the scope is thread-local, and a `#[tokio::test]` body (including `multi_thread`) is polled by `block_on` on the guard's thread, so straight-line code is captured across `.await`. Work handed to `tokio::spawn` / `spawn_blocking` is not — and that failure mode is loud, not silent: the expected series is simply absent.

## Evidence at default (parallel) concurrency

`djinn-supervisor --lib`, 6 consecutive runs:

| | baseline (`origin/main`) | this branch |
|---|---|---|
| run 1 | 6 failed | 189 passed |
| run 2 | 4 failed | 189 passed |
| run 3 | 6 failed | 189 passed |
| run 4 | 6 failed | 189 passed |
| run 5 | 7 failed | 189 passed |
| run 6 | 5 failed | 189 passed |

The failing set varied every baseline run, across all nine telemetry tests.

`djinn-coordinator --lib`: both `session_reaping` strike-decision tests failed on every baseline full-suite run and are green on this branch.

## Teeth checks

Assertions rewritten to absolute counts can accidentally assert nothing, so production emission was mutated and the tests re-run:

- `increment_strike_decision` given an extra `"task_id"` label → both coordinator tests fail.
- `increment_strike_decision` `.increment(1)` → `.increment(2)` → both fail with `left: 2.0 / right: 1.0`.
- `timed_clone_attempt` / `timed_cleanup_operation` record twice → **all nine** rewritten supervisor tests fail with `2.0 vs 1.0`.

All mutations reverted; the telemetry diff here is purely additive.

## Not in this PR

**Cluster B (`djinn-control-plane`, `doctor_closed_parent_tools.rs`)** was already fixed on `main` by #2820, which landed after the branch point this was reported from. Verified both directions: 8/8 green on current `main`; reverting #2820's five files reproduces it exactly (3 of 5 fail, varying set, `outcome: "skipped_not_found"`). The mechanism is *not* a shared test database — every `McpTestHarness` already gets its own ephemeral Postgres DB via `Database::open_in_memory()`. It was the process-global `djinn_core::doctor::registry()` singleton, keyed by check name, so concurrent harnesses clobbered each other's registration and ran each test's repair against a sibling's database. Nothing to do here.

**`rules::tests::batch_completion_creates_decomposition_task`** was reported alongside the coordinator metrics tests but is a **different mechanism** and is not fixed here. It has no metrics assertion; it fails its 10 s poll for a planning task that is never created (`left: 0`). It reproduces only under full-suite load (~25% of runs; 5/5 green alone, 3/3 green for the whole `rules` module) and it takes a sibling with it (`rules::tests::proposal_review_dispatched_incrementally_on_each_epic_close`). Failing-run logs show the test's own spawned coordinator *dispatching the fixture's second worker task into the live mock slot pool* moments before the test closes it, after which the actor emits nothing for the remaining 10 s. So the fixture races its own dispatch rather than the metrics registry. Fixing it means reworking the shared `spawn_coordinator` fixture in `rules.rs` on an unproven theory, in a file the follow-up `ubne`/`plcj` slices touch — reporting rather than racing, per instructions.

### DIFF
--- a/server/crates/djinn-coordinator/src/tests/session_reaping.rs
+++ b/server/crates/djinn-coordinator/src/tests/session_reaping.rs
@@ -14,11 +14,15 @@ fn rendered_counter_value(metric: &str, kind: &str) -> f64 {
         .unwrap_or(0.0)
 }
 
-/// Read one bounded strike-decision series from the process recorder. The
-/// cross-path tests take a render delta around the live coordinator call, not
-/// around the retry classifier, because the coordinator owns metric emission.
-fn rendered_strike_decision_value(decision: &str, source: &str) -> f64 {
-    let rendered = djinn_telemetry::render().unwrap();
+/// Read one bounded strike-decision series out of an already-rendered registry.
+///
+/// The cross-path tests render an [`djinn_telemetry::IsolatedRecorder`] scoped
+/// to the calling thread rather than the process recorder, so the value is
+/// exactly what the live coordinator call under test emitted — no delta, and
+/// nothing a concurrently running sibling test emitted. A series that is
+/// absent means the emission did not happen on this thread, which is a real
+/// failure rather than a zero.
+fn strike_decision_value(rendered: &str, decision: &str, source: &str) -> f64 {
     let prefix = format!(
         "djinn_dispatch_strike_decisions_total{{decision=\"{decision}\",source=\"{source}\"}} "
     );
@@ -33,8 +37,13 @@ fn rendered_strike_decision_value(decision: &str, source: &str) -> f64 {
 /// Guard the cardinality contract at the rendered recorder boundary. Dispatch
 /// identity is durable evidence and structured logging context, never a metric
 /// label: this counter may have exactly its bounded decision/source pair.
-fn assert_strike_decision_labels_are_bounded() {
-    let rendered = djinn_telemetry::render().unwrap();
+///
+/// `rendered` must come from the test's own [`djinn_telemetry::IsolatedRecorder`].
+/// Reading the process-global registry here made the guard assert on every
+/// series the *whole binary* had ever emitted, so a sibling test exercising a
+/// different — but equally bounded — source (`environmental_restart_orphan`,
+/// which this call path never emits) failed this test instead of its own.
+fn assert_strike_decision_labels_are_bounded(rendered: &str) {
     let allowed_decisions = ["counted", "exempted"];
     let allowed_sources = [
         "environmental_owner_expired",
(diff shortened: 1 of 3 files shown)


fernandobandeira:
Post-hoc evidence: `djinn-coordinator --lib`, 8 consecutive runs at default concurrency on this branch.

| run | result |
|---|---|
| 1 | 2132 passed |
| 2 | 2 failed — `rules::tests::batch_completion_creates_decomposition_task`, `rules::tests::proposal_review_dispatched_incrementally_on_each_epic_close` |
| 3 | hung on `rules::tests::batch_completion_does_not_fire_for_closed_epic` (SIGKILLed after 5 min) |
| 4 | 2132 passed |
| 5 | 2132 passed |
| 6 | 1 failed — `rules::tests::rapid_successive_building_amends_coalesce_to_one_open_reconcile_task` |
| 7 | 2132 passed |
| 8 | 1 failed — `wave::tests::blocked_epic_does_not_get_planning_task_while_blocker_open` |

**Zero `session_reaping` strike-decision failures across all 8 runs** — before this change both failed on every run.

Every residual failure is in the `rules::tests` / `wave::tests` families, a different test each time, all built on the same `spawn_coordinator` background-actor + bounded-wait fixture. Five distinct test names across four runs plus one hang: that is a third cluster with its own mechanism, unrelated to the metrics registry, and it is not addressed here (see "Not in this PR").

`Vercel` is red for an account-level reason — `Deployment rate limited — retry in 24 hours` — not anything in this diff. Every other check is green, including all eight Server Test shards, Server Clippy, and Server sqlx Freshness.

PR #2859, opened 2026-07-31
feat(3i92): stock-chart cutover preflight — fail closed on six defect classes (zpen)

fernandobandeira:
Implements board task `zpen` (epic `xowm`, proposal `3i92`): the startup/deploy preflight that must fail closed before a deployment is flipped from `leaf-v1` to `resize-v2`.

## What landed

| file | role |
|---|---|
| `server/crates/djinn-k8s/src/cutover_preflight.rs` | the real validator — one entry point, `run`, over six defect classes |
| `server/crates/djinn-k8s/src/bin/cutover-preflight.rs` | deploy-time driver, mirroring `bin/render-gate.rs` |
| `server/crates/djinn-k8s/tests/cutover_preflight.rs` | 25 proofs; 18 of them `#[ignore]`d because they need helm + Postgres |
| `deploy/preflight/cutover-preflight.sh` | renders the chart, extracts the rendered `DJINN_K8S_*`, `env -i`s into the binary |
| `deploy/preflight/tests/cutover-preflight.sh` | 14-case acceptance contract for that driver |
| `deploy/helm/djinn/tests/pods-resize-rbac.sh` | extended (additive) with the two chart↔preflight couplings |
| `.github/workflows/quality-gate.yml` | one new `cutover-preflight` job, wired into the aggregating gate |

### It calls the existing seams; it does not reimplement them

* Catalog/protocol → `djinn_db::launcher_compatibility::decide_admission` (the centralized admission decision, composing `z3gi`'s declaration verdict with the signed legacy-digest inventory). No second copy of "which images may run".
* Birth confirmation → `djinn_k8s::pod_resize::confirm_launcher_cpu`.
* Drain fence → `BuildPodPermitRepository::list_nonterminal_resize` (production query, production partial index).
* Ceiling → `rendered_lease_millicores` off the render, plus `apply_launcher_authority_protocol` in the driver.

### AC4 is protocol-conditional, in both directions

`launcher.rs` documents the absence of a launcher CPU limit under `leaf-v1` as deliberate: a limit there is an **ancestor clamp** over every invocation leaf, and task `7deu` measured a leaf set to 4 cores burning 0.25 with the leaf's own `nr_throttled` reading 0 because the throttling happened at the parent.

So the epic's unconditional wording is implemented conditionally, and both directions are asserted:

* `resize-v2` — absent ceiling FAILS, a ceiling below the 50m `LAUNCHER_CPU_REQUEST` FAILS, a ceiling that disagrees with the rendered lease FAILS.
* `leaf-v1` — that same absent ceiling **PASSES** (required), and a *present* one FAILS as the 7deu clamp.

The branch is `LauncherAuthorityProtocol::launcher_owns_leaf_quota()`, never a string; a source gate (`the_ceiling_branch_is_gated_on_the_predicate_not_a_protocol_string`) enforces that, because the two spellings are behaviourally identical *today* and stop agreeing the moment a third protocol lands.

The ceiling is read back off `DJINN_LAUNCHER_LEASED_MILLICORES` via `rendered_lease_millicores`, never recomputed from config — a per-project `build_resources.task.cpu_limit` override otherwise clamps a pod below its own lease.

---

## Mutation-verification record

Every run below is `cargo test -p djinn-k8s --test cutover_preflight -- --include-ignored`, real output, one mutation at a time, reverted after. Baseline: **25 passed; 0 failed**.

### 1. `pods-resize-rbac` — deleted the body of `check_pods_resize_rbac`

```
failures:
    a_cluster_wide_pods_resize_grant_is_refused
    each_pods_resize_rbac_mutation_is_refused_and_names_the_subresource

test result: FAILED. 22 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.68s
```

`each_pods_resize_rbac_mutation_is_refused_and_names_the_subresource` drives the four render fixtures the AC names: rule deleted, verb `patch`→`get`, apiGroup `""`→`apps`, resource `pods/resize`→`pods`. The verb case is why the triple is matched as a triple — with `get` the rule still *mentions* `pods/resize`.

### 2. `launcher-cpu-ceiling` — deleted the body of `check_launcher_ceiling`

```
failures:
    a_ceiling_below_the_launcher_cpu_request_is_refused_under_resize_v2
    a_present_launcher_ceiling_under_leaf_v1_is_refused_as_an_ancestor_clamp
    an_absent_launcher_ceiling_fails_under_resize_v2_and_passes_under_leaf_v1
    resize_v2_without_a_launcher_sidecar_is_refused_but_leaf_v1_is_not

test result: FAILED. 20 passed; 4 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.93s
```

### 2b. AC4's named mutation — made the ceiling check **unconditional** (`if mode.launcher_owns_leaf_quota()` → `if false`)

```
failures:
    a_present_launcher_ceiling_under_leaf_v1_is_refused_as_an_ancestor_clamp
    an_absent_launcher_ceiling_fails_under_resize_v2_and_passes_under_leaf_v1
    an_allowlisted_legacy_digest_blocks_a_resize_v2_cutover
    the_live_render_is_a_clean_cutover_under_both_protocols

test result: FAILED. 20 passed; 4 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.68s
```

The required-passing `leaf-v1` half is what goes red. An unconditional reading of AC4 cannot ship green.

### 3. `birth-confirmation` — deleted the body of `check_birth_confirmation`

```
failures:
    a_misleading_container_status_never_confirms_a_birth_downsize
    birth_confirmation_covers_missing_duplicate_stale_and_canonical_forms

test result: FAILED. 22 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.29s
```

### 3b. AC6's named mutation — `pod_resize::locate_launcher_status` reads `status.containerStatuses`

```
failures:
    a_ceiling_of_4_equals_an_expectation_of_4000_millicores
    a_misleading_container_status_never_confirms_a_birth_downsize
    birth_confirmation_covers_missing_duplicate_stale_and_canonical_forms

test result: FAILED. 21 passed; 3 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.49s
```

The misleading fixture is a Pod with a *matching* `status.containerStatuses[cgroup-launcher]` entry and no init-container status. A happy-path-only assertion survives this mutation; that fixture does not.

### 4. `catalog-protocol` — dropped the authority mode from the comparison

```
failures:
    an_allowlisted_legacy_digest_blocks_a_resize_v2_cutover
    the_catalog_truth_table_is_the_full_cross_product_of_both_enums

test result: FAILED. 22 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.29s
```

Exactly the cell AC7 names: (allowlisted-legacy-digest, `resize-v2`) flips to dispatchable and the literal truth table refuses it. The matrix is `CatalogDeclaration::ALL × LauncherAuthorityProtocol::ALL`, and the expectation table's row index is an exhaustive `match`, so a fifth variant fails to compile rather than skipping a cell.

### 4b. AC7's second mutation — `PreProtocolDigest::parse` accepts a mutable tag

```
thread 'a_mutable_tag_is_never_an_allowlisted_digest' panicked at crates/djinn-k8s/tests/cutover_preflight.rs:1018:5
failures:
    a_mutable_tag_is_never_an_allowlisted_digest

test result: FAILED. 24 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.31s
```

### 5. `drain-fence` — deleted the body of `check_drain_fence`

```
failures:
    an_unobservable_drain_fence_is_a_defect
    one_nonterminal_resize_row_in_any_state_blocks_the_cutover

test result: FAILED. 22 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.28s
```

### 5b. AC8's named mutation — narrowed `NONTERMINAL_RESIZE_STATES` by one state (dropped `birth_confirmed`)

```
failures:
    one_nonterminal_resize_row_in_any_state_blocks_the_cutover

test result: FAILED. 23 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.34s
```

That proof seeds ONE permit per nonterminal state, each in its own ephemeral template-cloned Postgres, walking the real lifecycle (`acquire` → `bind_or_refresh_job_uid` → `capture_resize_identity` → fenced transitions). No raw SQL leaves `djinn-db`.

### 5c. AC8's source gate — substituted a `Fake` repository

```
thread 'the_drain_fence_proofs_use_a_real_pg_pool_and_no_fake_repository' panicked at crates/djinn-k8s/tests/cutover_preflight.rs:1321:9
failures:
    the_drain_fence_proofs_use_a_real_pg_pool_and_no_fake_repository

test result: FAILED. 24 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.31s
```

### 6. `credential-boundary` — deleted the body of `check_credential_boundary`

```
failures:
    each_credential_boundary_mutation_is_refused

test result: FAILED. 23 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.45s
```

That proof carries all three of AC9's mutations, on the rendered artifact: `automountServiceAccountToken` set to `Some(true)` **and** removed; the projected token audience changed to `kubernetes.default.svc` **and** removed (absent = the apiserver's own default audience); and a `RoleBinding` naming the task-run ServiceAccount added to the live render.

### AC1 — replaced the whole body of `run` with `Ok(())`

```
failures:
    a_ceiling_below_the_launcher_cpu_request_is_refused_under_resize_v2
    a_cluster_wide_pods_resize_grant_is_refused
    a_misleading_container_status_never_confirms_a_birth_downsize
    a_present_launcher_ceiling_under_leaf_v1_is_refused_as_an_ancestor_clamp
    an_absent_launcher_ceiling_fails_under_resize_v2_and_passes_under_leaf_v1
    an_allowlisted_legacy_digest_blocks_a_resize_v2_cutover
    an_unobservable_drain_fence_is_a_defect
    birth_confirmation_covers_missing_duplicate_stale_and_canonical_forms
    each_credential_boundary_mutation_is_refused
    each_pods_resize_rbac_mutation_is_refused_and_names_the_subresource
    one_nonterminal_resize_row_in_any_state_blocks_the_cutover
    resize_v2_without_a_launcher_sidecar_is_refused_but_leaf_v1_is_not

test result: FAILED. 12 passed; 12 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.33s
```

12 distinct named tests, covering all six classes. AC1 asks for at least six.

### AC5 — replaced both millicore comparisons with string equality

Ceiling arm: `u64::from(lease) == ceiling.millis()` → `rendered == format!("{lease}m")`. Birth arm: `observed == target` → `raw == target.as_quantity()`.

```
failures:
    a_ceiling_of_4_equals_an_expectation_of_4000_millicores

test result: FAILED. 23 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.46s
```

That case renders the sidecar ceiling as `4` (what the apiserver stores a `4000m` PATCH back as, and the repo's own stock worker `cpu_limit`) against a `4000m` expectation and a `4000m` birth target.

### AC10 — the driver exits 0 unconditionally

Shell wrapper replaced with `echo CLEAR; exit 0`:

```
FAIL [stock chart defaults are a clean leaf-v1 deployment]: expected exit 1, got 0
FAIL [an armed chart is a clean leaf-v1 deployment]: expected exit 1, got 0
FAIL [an armed chart is ready for the resize-v2 cutover]: expected exit 1, got 0
FAIL [resize-v2 with the launcher disabled is refused]: expected exit 1, got 0
FAIL [render mutation rbac-verb is refused]: expected exit 1, got 0
FAIL [render mutation rbac-group is refused]: expected exit 1, got 0
FAIL [render mutation rbac-resource is refused]: expected exit 1, got 0
FAIL [render mutation rbac-deleted is refused]: expected exit 1, got 0
FAIL [a RoleBinding naming the task-run ServiceAccount is refused]: expected exit 1, got 0
FAIL [a pre-handshake catalog image blocks a resize-v2 cutover]: expected exit 1, got 0
FAIL [a matching status.containerStatuses entry never confirms a birth]: expected exit 1, got 0
FAIL [a ceiling of 4 confirms a target of 4000m]: expected exit 1, got 0
FAIL [a malformed authority mode is unevaluable]: expected exit 2, got 0
FAIL: 13 of 14 cutover-preflight cases failed
```

Same result with the *binary* mutated to `ExitCode::SUCCESS` unconditionally (13/14 red), so the contract binds the Rust exit path and not only the wrapper.

### AC2 — pointed the renderer at a static checked-in file

**This mutation initially stayed green, and the suite was strengthened until it did not.** A stale known-good render passes a diff-based guard forever, because a stale fixture still diffs cleanly against itself. So `the_render_under_test_is_produced_by_helm_at_test_time` now copies the chart to a temp dir, writes a fresh UUID nonce into its `values.yaml`, renders it through the SAME `render_chart` primitive `live_render` uses, and requires the nonce to come back out:

```
---- the_render_under_test_is_produced_by_helm_at_test_time stdout ----
thread panicked at crates/djinn-k8s/tests/cutover_preflight.rs:180:5:
the nonce "zpen-019fb8850fb47f83a5ba1f51adc71f2f" did not reach the render, so these documents
were not produced by helm from the chart on disk. Rendered names: ["djinn",
"djinn-cutover-preflight-djinn-controller", ...]

failures:
    the_render_under_test_is_produced_by_helm_at_test_time

test result: FAILED. 24 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.29s
```

`each_negative_fixture_differs_from_the_live_render_in_exactly_one_path` covers the other half: each negative fixture is fingerprinted against the live render, and RBAC rules are fingerprinted by *content* rather than by array index, so deleting one rule is exactly one removal instead of an index shift across every rule after it.

---

## Wiring, and why the `#[ignore]`s are safe

18 of the 25 proofs are `#[ignore]`d: they need `helm` (for the live render every fixture derives from) and real Postgres (for the drain fence), and the sharded lane has only the latter. An `#[ignore]` nobody runs is precisely the "merged, green, unreachable" defect this task exists to remove, so three always-on guards close it:

* `the_cutover_preflight_lane_is_wired_and_cannot_silently_skip` reads `.github/workflows/quality-gate.yml` back and fails if the job, its `--run-ignored all` invocation, the helm/Postgres setup, the `DJINN_DATABASE_URL` for the driver, or the declared proof count disappears — **and** if the aggregating `quality-gate` job stops `needs:`ing and `check`ing this lane.
* The declared count (`DJINN_CUTOVER_EXPECTED_PROOFS: "18"`) is compared against the `#[ignore]` count in the file, both ways.
* The lane itself fails if fewer than that many tests actually execute.

`deploy/preflight/tests/cutover-preflight.sh` is also globbed by the existing `deploy-render-gate` lane (`scripts/test-deploy-contracts.sh deploy/preflight/tests`), where there is no database — so it asserts the exact defect *class set* rather than only the exit status, proving every render-derived class came back clean even when the drain fence is legitimately UNOBSERVABLE.

## Chart↔preflight couplings

`deploy/helm/djinn/tests/pods-resize-rbac.sh` gains two additive checks for the ways the preflight could pass *vacuously* if the chart drifted: the triple it matches must be the triple this render grants (proved non-vacuous against drifted apiGroup and verb constants), and the `app.kubernetes.io/component: taskrun` label it keys on to name the ServiceAccount must exist and be unique (proved non-vacuous by stripping it).

The resource half is asserted as "the constant names a `pods/` subresource" rather than as a render match, and the script says why: the chart legitimately grants `resources: ["pods"]` on the line above, so a constant drifting to `pods` would still find exactly one matching Role rule. Claiming otherwise would have been the vacuous assertion that file exists to forbid.

## The lane's first run found a real defect in the lane

Run 30638114402 executed all 25 proofs against helm v3.12.3 and Postgres 16 and every one passed — then the job failed anyway:

```
     Summary [   2.533s] 25 tests run: 25 passed, 0 skipped
::notice::cutover-preflight proofs executed=0 expected>=18
::error::executed 0, expected at least 18. An ignored proof nobody runs is the exact defect this lane exists to prevent.
```

This workflow sets `CARGO_TERM_COLOR: always`, so nextest colorizes even into a pipe and every line is `ESC[32;1m        PASS ESC[0m …`. The counter matched raw text and read 0. Worth stating plainly: the guard fails **closed** on a count it cannot read, which is why a formatting mismatch surfaced as a red lane instead of a silently unenforced one. Fixed by stripping ANSI before counting, plus an explicit `0 skipped` assertion.

## …and then found a real defect in the driver

With the counter fixed, run 30638423759 ran all 25 Rust proofs green and the driver contract failed — with a real bug, on the only configuration that can reach it:

```
note: DJINN_DATABASE_URL is set; the drain fence is read for real
FAIL [stock chart defaults are a clean leaf-v1 deployment]: expected exit 0, got 101
thread 'main' panicked at sqlx-core-0.8.6/src/pool/inner.rs:529:5:
this functionality requires a Tokio context
```

`Database::open_with_config` was called beside the runtime rather than inside it. `PgPool` spawns its own reaper at construction, so sqlx panics — and a panicking deploy gate exits 101, which reads as neither a clean nor a blocked verdict. Every local run had `DJINN_DATABASE_URL` unset, so the path was never taken until the lane supplied one. Fixed by moving the pool construction inside `block_on`.

Re-verified locally against a migrated Postgres (a `djinn_test_template` clone), which is the first configuration in which a clean verdict is a genuine exit 0:

```
note: DJINN_DATABASE_URL is set; the drain fence is read for real
ok [stock chart defaults are a clean leaf-v1 deployment] classes={}
ok [an armed chart is a clean leaf-v1 deployment] classes={}
ok [an armed chart is ready for the resize-v2 cutover] classes={}
ok [resize-v2 with the launcher disabled is refused] classes={launcher-cpu-ceiling}
ok [render mutation rbac-verb is refused] classes={pods-resize-rbac}
ok [render mutation rbac-group is refused] classes={pods-resize-rbac}
ok [render mutation rbac-resource is refused] classes={pods-resize-rbac}
ok [render mutation rbac-deleted is refused] classes={pods-resize-rbac}
ok [a RoleBinding naming the task-run ServiceAccount is refused] classes={credential-boundary}
ok [the verdicts above came from the render; the arguments never changed]
ok [a pre-handshake catalog image blocks a resize-v2 cutover] classes={catalog-protocol}
ok [a matching status.containerStatuses entry never confirms a birth] classes={birth-confirmation}
ok [a ceiling of 4 confirms a target of 4000m] classes={}
ok [a malformed authority mode is unevaluable, not defaulted to leaf-v1]
PASS: cutover preflight contract (14 cases)
```

Both defects were found by the lane and the contract this PR adds, on their first two runs. That is the argument for the lane.

## Notes

* No live cluster is used or created.
* `check-raw-sql-boundary.sh`: clean — the drain-fence fixtures go through `djinn-db` repositories only.
* `check-file-size.sh`: the test file carries `djinn:allow-oversize` with a stated reason.
* `cargo clippy -p djinn-k8s --all-targets` and `cargo fmt --check`: clean.

### DIFF
--- a/.github/workflows/quality-gate.yml
+++ b/.github/workflows/quality-gate.yml
@@ -913,6 +913,128 @@ jobs:
         run: bash -c "$CI_RUNNER_DISK_PREFLIGHT"
       - name: Test deploy render-gate contract
         run: bash scripts/test-deploy-contracts.sh deploy/preflight/tests
+  cutover-preflight:
+    name: Cutover Preflight
+    needs: preflight
+    # Same routing as `deploy-render-gate`: this contract's inputs are the chart
+    # and the djinn-k8s crate, and every path that can change either lands in
+    # the router's `rustCore` lane or fails closed into `unknown`.
+    if: needs.preflight.outputs.clippy == 'true' && github.event_name != 'push'
+    runs-on: ubuntu-latest
+    # The deploy contract builds one binary and shells out to helm; the Rust
+    # proofs need a migrated Postgres template. 30 leaves margin for a cold
+    # djinn-k8s chain without hiding a hang.
+    timeout-minutes: 30
+    defaults:
+      run:
+        working-directory: server
+    env:
+      SQLX_OFFLINE: "true"
+      # Number of `#[ignore]`d proofs `tests/cutover_preflight.rs` declares.
+      #
+      # Those proofs are ignored because they need `helm` (for the live chart
+      # render every fixture is derived from) and a real Postgres (for the drain
+      # fence) — neither of which the ordinary sharded lane has. An `#[ignore]`
+      # nobody runs is precisely the "merged, green, unreachable" defect this
+      # task exists to remove, so the run below fails if fewer than this many
+      # actually execute, and the suite's always-on
+      # `the_cutover_preflight_lane_is_wired_and_cannot_silently_skip` reads
+      # this file back and fails if this declaration, the job, or the
+      # `--run-ignored all` invocation disappears.
+      DJINN_CUTOVER_EXPECTED_PROOFS: "17"
+      # The drain-fence proofs open template-cloned per-test databases here.
+      DJINN_TEST_DATABASE_URL: postgres://postgres:postgres@127.0.0.1:5433/postgres
+    services:
+      postgres:
+        image: postgres:16
+        env:
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: djinn
+        ports:
+          - 5433:5432
+        options: --health-cmd="pg_isready -U postgres" --health-interval=5s --health-timeout=3s --health-retries=20
+    steps:
+      - uses: actions/checkout@v4
+      # Pinned to the same v3.12.3 deploy-render-gate installs — two lanes
+      # rendering the same chart on different helm majors is a contract that
+      # means nothing.
+      - uses: azure/setup-helm@v4
+        with:
+          version: v3.12.3
+      - run: rustup toolchain install
+        working-directory: server
+      - uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: server
+          shared-key: server-test
+          cache-on-failure: true
+          cache-all-crates: true
+          save-if: false
+        id: rust-cache
+      - name: Log cache family and restore status
+        run: |-
+          echo "cache-family=server-test"
+          echo "shared-key=server-test"
+          echo "runner-os=${{ runner.os }}"
+          echo "rustc=$(rustc --version || echo 'unknown')"
+          echo "cache-hit=${{ steps.rust-cache.outputs.cache-hit }}"
+          echo "::notice::cache-family=server-test shared-key=server-test cache-hit=${{ steps.rust-cache.outputs.cache-hit }}"
+      - name: Runner disk headroom preflight
+        # Body/thresholds: CI_RUNNER_DISK_PREFLIGHT at the top of this file.
+        run: bash -c "$CI_RUNNER_DISK_PREFLIGHT"
+      - name: Apply migrations
+        env:
+          DATABASE_URL: postgres://postgres:postgres@127.0.0.1:5433/djinn
+        run: |
+          until pg_isready -h 127.0.0.1 -p 5433 -U postgres >/dev/null 2>&1; do sleep 1; done
+          SQLX_OFFLINE=true \
+            DJINN_MIGRATE_DATABASE_URL=postgres://postgres:postgres@127.0.0.1:5433/djinn \
+            DJINN_MIGRATE_OPERATOR_USER_ID=00000000-0000-7000-8000-000000000002 \
+            DJINN_MIGRATE_OPERATOR_GITHUB_ID=9000000002 \
+            DJINN_MIGRATE_OPERATOR_LOGIN=djinn-ci-migration-operator \
+            cargo test -p djinn-db --test apply_migrations -- --ignored --nocapture
+      - name: Build Postgres test template
+        env:
+          PGPASSWORD: postgres
+        run: |
+          psql -h 127.0.0.1 -p 5433 -U postgres -d postgres -v ON_ERROR_STOP=1 -c "DROP DATABASE IF EXISTS djinn_test_template"
+          psql -h 127.0.0.1 -p 5433 -U postgres -d postgres -v ON_ERROR_STOP=1 -c "CREATE DATABASE djinn_test_template"
+          SQLX_OFFLINE=true DJINN_TEST_DATABASE_URL=postgres://postgres:postgres@127.0.0.1:5433/postgres cargo test -p djinn-db --test setup_test_template -- --ignored
+          psql -h 127.0.0.1 -p 5433 -U postgres -d postgres -v ON_ERROR_STOP=1 -c "UPDATE pg_database SET datistemplate = TRUE WHERE datname = 'djinn_test_template'"
+      - uses: taiki-e/install-action@nextest
+      - name: Prove the cutover preflight against a live render and real Postgres
+        run: |
+          set -euo pipefail
+          # `--run-ignored all` runs the ignored proofs ALONGSIDE the always-on
+          # guards, so the wiring guard and the source gates are covered by the
+          # same invocation. The executed count is compared against the
+          # declaration above: a lane that quietly runs zero proofs must fail
+          # rather than look green.
+          cargo nextest run -p djinn-k8s --test cutover_preflight \
+            --run-ignored all --no-fail-fast --message-format libtest-json \
+            > "$RUNNER_TEMP/cutover.json" 2> "$RUNNER_TEMP/cutover.log" || {
+            cat "$RUNNER_TEMP/cutover.log"
+            exit 1
+          }
+          cat "$RUNNER_TEMP/cutover.log"
+          executed=$(grep -c '"type": *"test", *"event": *"ok"' "$RUNNER_TEMP/cutover.json" || true)
+          echo "::notice::cutover-preflight proofs executed=$executed expected>=$DJINN_CUTOVER_EXPECTED_PROOFS"
+          if [ "$executed" -lt "$DJINN_CUTOVER_EXPECTED_PROOFS" ]; then
+            echo "::error title=Cutover preflight proofs did not run::executed $executed, expected at least $DJINN_CUTOVER_EXPECTED_PROOFS. An ignored proof nobody runs is the exact defect this lane exists to prevent."
+            exit 1
+          fi
+        env:
+          NEXTEST_EXPERIMENTAL_LIBTEST_JSON: "1"
+      - name: Test the deploy-time cutover preflight driver
+        # The exit code of deploy/preflight/cutover-preflight.sh is the whole
+        # contract, and this suite is what asserts it. DJINN_DATABASE_URL is
+        # supplied so the drain fence is READ rather than reported unobservable,
+        # which is the only configuration in which a clean verdict is a genuine
+        # exit 0.
+        working-directory: ${{ github.workspace }}
+        env:
+          DJINN_DATABASE_URL: postgres://postgres:postgres@127.0.0.1:5433/djinn
+        run: bash deploy/preflight/tests/cutover-preflight.sh
   launcher-kernel-boundary:
     name: Launcher Kernel Boundary
     needs: preflight
@@ -2301,6 +2423,7 @@ jobs:
       - server-guards
       - server-clippy
       - deploy-render-gate
+      - cutover-preflight
       - server-aarch64-check
       - nextest-plan
       - server-test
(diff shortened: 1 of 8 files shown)


fernandobandeira:
CI is green on `47884c809` (run 30638828922): **27 pass, 3 skipping** (the deliberately-skipped cache-warm lanes), Quality Gate success.

One note for the reviewer, so the run history is not misread. On its first attempt this run showed all 8 `Server Test` shards red. They were **cancelled**, not failed — every lane in the run concluded `success`, and the aggregating gate's only complaint was `lane server-test-shards selected=true result=cancelled`. `gh run rerun --failed` returned all 8 green with no code change. Attribution: not this PR. The shards' fail-fast watcher only fires on a concluded `failure`, and there was none, so the run-level cancel came from outside it — the same class the runner-disk preflight comment at the top of `quality-gate.yml` documents.

The two defects this PR's own lane caught on its first two runs (an ANSI-blinded proof counter, then a `PgPool` built outside the Tokio runtime) are both recorded in the PR body with their real output.

PR #2864, opened 2026-07-31
feat(resize): status-confirmed lift — arm ResizeAuthority in AppState behind a UID/restart/protocol-fenced PATCH-and-poll (0ppk-1c, 0vku)

fernandobandeira:
Board task `0vku` (0ppk-1c, "the lift") of epic `0ppk`, proposal `3i92`. Builds directly on #2860 (`pwys`, 0ppk-1b), which merged.

## The contradiction this PR exists to fix

#2845 shipped the resize authorization seam as:

```rust
pub trait PodResizeIntentSink: Send + Sync {
    fn record(&self, intent: &PodResizeIntent);   // sync, infallible, returns ()
}
```

and `fold_into_grant` returned `Authorized(_) => granted`, dropping the authorized intent on the floor.

An async `pods/resize` PATCH plus a status-confirmation poll **cannot report failure through that trait**. Plugged in as-is it would have been a lift that structurally cannot fail: every uncertainty test would pass without exercising anything, and this epic would collect its eighth green, merged, inert slice. So the seam changed:

```rust
#[async_trait]
pub trait PodResizeApplier: Send + Sync {
    async fn apply(&self, intent: &PodResizeIntent) -> Result<(), ResizeApplyFailure>;
}
```

and the `Authorized` arm now awaits the apply and maps failure onto `LeaseResult::DegradedUnleased`.

**Reverting the seam to `fn record(&self, &PodResizeIntent)` is the named mutation.** It does not merely fail a test — `resize_lift.rs` and every uncertainty case in `resize_lift_tests.rs` stop compiling, because there is nowhere left to put the failure.

## What is new

**`server/crates/djinn-coordinator/src/resize_lift.rs`** — the fenced apply-and-poll. `PodResizeClient::resize_launcher_cpu(pod_name, target)` takes only a Pod *name*; from that signature alone four fences are missing, and all four are supplied here:

* **UID fence** — the live `metadata.uid` is compared against the permit's write-once `pod_uid` on the fresh GET *before* any PATCH, and again after confirmation. A Pod deleted and recreated under the same name is refused with the PATCH counter at **zero**.
* **Restart fence** — the permit's captured `launcher_container_id` vs the live one.
* **Protocol fence** — the permit's `effective_launcher_protocol` vs the launcher's declared `DJINN_LAUNCHER_AUTHORITY_PROTOCOL`.
* **Bounded confirmation poll** — the client does one GET/PATCH/GET cycle and no retries; `status.initContainerStatuses` is populated asynchronously.

Every fence re-runs on every poll pass, so a Pod replaced or a launcher restarted *during* confirmation stops the lift rather than letting the next PATCH land on a different object.

Reused verbatim, never re-derived: `Patch::Strategic`, the refusal to consult `spec.containers` / `status.containerStatuses`, the `PodResizePending` fail-closed, and the **millicore** comparison.

**`DegradedUnleasedReason` gains 14 apply-time variants** — a closed set on `DegradedUnleased`, never an `Err` the invocation runner would treat as retryable. Each gets its own metric label in `degrade_reason`.

**A failed lift leaves the permit `drop_required`.** `birth_confirmed → lift_applying` before the PATCH; `→ lifted` on confirmation; `→ drop_required` on any failure. Migration 164's trigger owns the legal edge list — no migration is added by this PR.

**`PodResizeError::Api` gains `status: Option<u16>`**, populated from `kube::Error::Api(ErrorResponse { code, .. })`. 403 (missing `pods/resize` RBAC), 422 (the apiserver refused this resize) and a transport timeout (nothing known) are three different operator facts, and recovering that distinction by matching a rendered error string would make the classification depend on `kube`'s `Display` impl.

## Arming

`AppState::new_inner` now calls `.with_resize_authority(...)`. That call had **zero production call sites** for the whole of 0ppk-1a.

## Inert on the current fleet today

`images.launcher_authority_protocol` is NULL fleet-wide, so every dispatch resolves `leaf-v1`. Under `leaf-v1` no launcher `limits.cpu` is rendered, no resize identity is captured (migration 164 requires `admitted_cpu_millicores > 0`), and `clamp` refuses non-`resize-v2` outright with `ProtocolNotResizable`. **Nothing in this PR moves a Pod in production as it stands.**

What makes it live: setting `launcher_authority_protocol = 'resize-v2'` on an image row. From that point a `resize-v2` dispatch captures a ceiling, birth-downsizes to 250m, and an invocation escalation lifts it back — status-confirmed or degraded.

## The second composition site: `DirectServices` fails closed

`DirectServices::with_provider_override` builds a second `BuildLeaseService` (cap 0) as an agent-side fallback. **It stays unarmed, deliberately**, asserted by `the_agent_side_fallback_build_lease_fails_closed_unarmed`:

1. `djinn-agent` cannot depend on `djinn-server` — that is *why* #2860 introduced the `TaskRunResizeAdmission` trait. The apiserver surface and the permit-creating dispatch seam are both on the far side.
2. It is not on the path that creates permits; arming it would answer `PermitAbsent` for every invocation it served.
3. Unarmed means "no lift" — the launcher keeps its birth quota. The unsafe direction is *reporting* a grant for CPU no Pod received, which is exactly what an armed-but-surfaceless composition does.

## CI

`check-resize-authorization-boundary.sh` now also requires a **production** caller of `with_resize_authority` under `server/src`. Excluded: `#[cfg(test)]` modules, `*_tests.rs`, `.worktrees/`, and — found while building it, because the first version passed on one — **commented-out calls**, line and block. Self-test extended in lockstep, 14 assertions, all watched failing.

## Named mutations, each RUN and observed failing

| Mutation | Result |
|---|---|
| Restore `Authorized(_) => granted` in `fold_into_grant` | 10 tests fail; happy-path PATCH counter goes to 0 |
| Remove the `observed.pod_uid != intent.pod_uid` comparison | recreated-Pod test grants, PATCH count 1 |
| Delete `.min(ceiling)` in `clamp` | `ZERO PATCH bodies above the admitted ceiling; observed [4000]` |
| Ceiling := `configured_leased_millicores` instead of the row | carried ceiling reads 4000, want 8000 |
| Delete `.with_resize_authority(...)` from `new_inner` | AC1 test fails **and** the CI guard fails naming the symbol |
| Add `.with_resize_authority(..)` to `DirectServices` | fail-closed test fails |

## Evidence discipline

The first version of the uncertainty table **passed vacuously** and I caught it: the fixture Pod still held its admitted ceiling, so the lift's target was a value the launcher already had and a "never actuates" node still confirmed. The fixture now performs the real birth downsize to 250m between identity capture and lift, and asserts the launcher reads 250m before any lift runs. `LiftStatusStale` and `a_matching_regular_container_status_is_never_confirmation` only became real assertions after that.

No `Fake`/`Mock` repository. Permits, the lease FIFO and the lift authority are the production types over `open_in_memory` (a template-cloned real Postgres). The single double is the HTTP transport, via `djinn_k8s::pod_resize_fixture` — whose observation, PATCH and confirmation are the production functions.

## What I could NOT satisfy — stated, not weakened

**1. AC6 and AC7 are mutually contradictory as written, and I did not redesign the clamp to hide it.**

AC6 requires configuring a lift *above* the captured ceiling and landing on the ceiling — that is `min(configured, admitted)`, what 0ppk-1a shipped. AC7 requires a per-project-override Pod admitted *above* the deployment default to lift to *its own* rendered lease — that is `admitted`, i.e. a `max` in that direction. One `min` cannot do both.

I kept the clamp exactly as 0ppk-1a shipped it and proved the half that is provable: AC7's **ceiling provenance** (the bound is the write-once captured value, and the named mutation makes it read 4000 instead of 8000). The behavioural half is **not met**: an override Pod admitted at 8000m still lifts only to the deployment default 4000m. Making it lift to 8000 means `target = admitted`, which satisfies both ACs — AC6's "configured above the ceiling" then lands on the ceiling by construction, and AC6's stated mutation still fires. I believe that is the right follow-up, but it changes a merged slice's semantics and belongs in its own PR with its own review, not smuggled in here.

**2. The live-kind-cluster halves are not done.** I was instructed not to create a cluster. Specifically missing:
* AC6's companion control — issuing the oversubscribed value *directly* through `pods/resize` against a live apiserver to prove the apiserver does not bound it.
* AC8's 20 lift/drop cycles on a real kubelet. The cluster-free equivalent runs 20 cycles through the production patch-body builder and strategic-merge semantics and asserts requests/QoS/containerID/restartCount/init-container-count byte-identical.

**3. Two AC4 cases share `LiftStatusStale`.** The plain-stale case and the misleading-regular-container case, because the launcher's own init status is stale in both — and telling them apart would require reading `status.containerStatuses`, the exact thing this stack is forbidden to do. Encoded as a stated decision in `the_apply_time_reasons_are_a_closed_distinct_set` rather than left to look accidental. The other nine cases have distinct reasons.

**4. `LiftDeadlineExceeded` vs the specific status reasons** are separated by whether waiting was permitted: a zero-budget lift reports what its one observation saw; a lift that slept its whole budget reports that it spent it. Both degrade. Documented on the variant.

**Unrelated pre-existing failure:** `djinn-coordinator`'s `rules::tests::amend_while_building_dispatches_one_reconcile_task_from_event` stack-overflows. Verified it does so on `origin/main` (134f35e0d) untouched, so it is not from this PR.

### DIFF
--- a/scripts/check-resize-authorization-boundary.sh
+++ b/scripts/check-resize-authorization-boundary.sh
@@ -25,6 +25,23 @@
 #      it is a comment that invites a reader to read it.
 #   3. Each required type is still named by the module. Deleting the predicate
 #      would otherwise satisfy check 2 perfectly.
+#   4. (0ppk-1c) `with_resize_authority` has at least one PRODUCTION caller.
+#
+# WHY CHECK 4
+#
+# `BuildLeaseService` holds its resize authorization as
+# `Option<Arc<ResizeAuthority>>`. For the whole of `0ppk-1a` that `Option` was
+# `None` at every composition and `with_resize_authority` had zero call sites
+# outside tests: the authorization layer, the clamp, and their entire test suite
+# were merged, green, and structurally unable to move a Pod. Nothing about that
+# produces a compile error, and no unit test can catch it — a unit test that
+# installs the authority itself proves only that the setter works.
+#
+# Reachability is a property of the COMPOSITION SITE, so it is checked here, by
+# text, over the production tree. `#[cfg(test)]` blocks, `*_tests.rs` files and
+# `.worktrees/` scratch copies are excluded deliberately: a caller in any of
+# those is exactly the false positive that would let the arming be deleted while
+# CI stayed green.
 #
 # Usage:
 #   ./scripts/check-resize-authorization-boundary.sh
@@ -42,6 +59,8 @@ cd "$ROOT"
 GUARDED_FILES="
 server/crates/djinn-coordinator/src/resize_authorization.rs
 server/crates/djinn-coordinator/src/resize_authorization_tests.rs
+server/crates/djinn-coordinator/src/resize_lift.rs
+server/crates/djinn-coordinator/src/resize_lift_tests.rs
 "
 
 # The retired handoff relation and the protocol columns flc5 drops with it.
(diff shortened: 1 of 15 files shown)


fernandobandeira:
Pre-existing baseline note, verified on `origin/main` (134f35e0d) with this branch's changes absent:

- `djinn-coordinator --lib rules::tests::amend_while_building_dispatches_one_reconcile_task_from_event`
- `djinn-server --lib server::tests::debug::test_debug_dispatch_state_does_not_mutate`

Both stack-overflow under a plain `cargo test` (default 2 MiB thread stack). Both overflow identically on untouched `main`, and CI is green because the shards run under nextest, which gives each test its own process. Recording it so the next person who runs `cargo test -p djinn-server --lib` locally does not spend an hour blaming this PR.

PR #2875, opened 2026-07-31
feat(images): make the declared launcher authority protocol configurable, and make a cached protocol flip impossible

fernandobandeira:
## The problem

`dockerfile.rs:52` declared the launcher authority protocol with a hardcoded constant:

```rust
pub const DECLARED_LAUNCHER_PROTOCOL: LauncherAuthorityProtocol = LauncherAuthorityProtocol::LeafV1;
```

There was no config, env var, or override. Every artifact this repo builds declared `leaf-v1`, so `images.launcher_authority_protocol` is `leaf-v1`/NULL fleet-wide and `render_authority_protocol` always resolves `LeafV1`. **3i92's resize arm could not be entered from any deployment**, and step 4 of `docs/deploy/launcher-authority-cutover.md` ("Rebuild and catalog images as `resize-v2`") had no mechanism behind it.

## The change

The declaration is now a **build input**, threaded from the deployment:

```
imagePipeline.controller.launcherAuthorityProtocol: "resize-v2"   # default "" → leaf-v1
  → DJINN_IMAGE_LAUNCHER_AUTHORITY_PROTOCOL (server Pod env)
  → ImageControllerConfig::declared_launcher_protocol
  → controller::render_build_context
  → generate_dockerfile(cfg, agent_worker, protocol)
  → LABEL djinn.app/launcher-authority-protocol + ENV DJINN_LAUNCHER_AUTHORITY_PROTOCOL
  → build Job env → DJINN_LAUNCHER_PROTOCOL= sentinel → images.launcher_authority_protocol
```

`emit_launcher_protocol` still writes exactly what was built, and the declaration still travels in build metadata — never inferred from a tag.

### Why deployment-level, not per project

The declaration describes the `djinn-cgroup-launcher` binary copied out of the **single** `agent_worker_image` every catalog image shares, and admission compares it against the cluster-wide `launcher_authority_mode` singleton. A per-project protocol would mean two projects on one cluster disagreeing about who owns a node's CPU quota — not something the mode row can express, and not something the cutover sequence (pause → drain → CAS the mode) can stage. One knob, one fleet.

## The trap, closed by construction

The old doc comment asked for a manual `compute_environment_hash` salt bump on every flip, because the hash did not cover the Dockerfile text. That is exactly the "cached image with the old launcher, catalogued as `resize-v2`" outcome. This PR removes the need for the discipline instead of documenting it harder:

1. **The protocol is an input to the environment hash**, and the image tag is a prefix of that hash. `leaf-v1` and `resize-v2` builds of one config land on **different tags** — a flip cannot resolve to the cached artifact and cannot relabel one in place. (The `leaf-v1` tags stay in the registry, which is what makes the runbook's retention check and rollback possible.)
2. **The hash is taken from the rendered context, not from a second restatement.** `enqueue_image` now renders the Dockerfile *first* and calls `BuildContext::environment_hash(...)`, which hashes `self.launcher_protocol`. There is no call site where the hashed declaration and the built one can drift, because there is only one value.
3. **`build_image_build_job` is fail-closed.** It calls `BuildContext::verify_declaration()` and returns `Err(DeclarationError)` unless the Dockerfile declares exactly what the context reports (also catching a missing declaration, a LABEL/ENV split, an unrecognised wire form, and a second declaration appended later). That Job's env becomes the catalog row and its Dockerfile becomes the artifact's label, so **an image whose label and catalog row disagree cannot be built** — it is not detected after the fact, it never gets a Job.

### Default is a true no-op

`DEFAULT_LAUNCHER_PROTOCOL` contributes **nothing** to the hash pre-image, so an unconfigured deployment computes byte-identical hashes to before this PR. Mixing it in unconditionally would have re-tagged and rebuilt every catalog image on every existing deployment at upgrade, for a change that alters no artifact. The golden Dockerfile fixtures are untouched, and the chart renders no env var unless the operator sets one.

## Acceptance criteria, each proven by a mutation I ran

| AC | Test | Mutation applied | Result |
|---|---|---|---|
| 1. A deployment can configure the protocol; the built artifact's metadata carries it | `a_configured_protocol_reaches_the_artifact_and_the_catalog_row`, `the_configured_protocol_is_what_the_artifact_declares`, `configuring_resize_v2_changes_only_the_declaration` | `render_build_context` passes `DEFAULT_LAUNCHER_PROTOCOL` instead of the config field (i.e. hardcoded again) | 2 tests FAIL |
| 2. A protocol change invalidates the image cache | `hash_changes_when_the_declared_protocol_changes`, `flipping_the_deployment_selection_invalidates_the_cached_image` | delete the `launcher_protocol` block from `compute_environment_hash` | 2 tests FAIL |
| 3. Default unchanged; no existing deployment rebuilds | `the_default_protocol_hashes_to_the_pre_change_preimage` (reconstructs the pre-`launcher_protocol` pre-image from its own bytes), golden fixtures byte-identical, every pre-existing assertion kept | hash the protocol unconditionally | 1 test FAILS |
| 4. Catalog row and artifact declaration cannot disagree | `a_build_whose_label_and_catalog_row_would_disagree_is_refused`, `a_context_that_reports_something_other_than_it_built_is_refused` | remove `verify_declaration()?` from `build_image_build_job` | test FAILS (the mismatched Job renders) |
| — chart reachability | `every_image_controller_env_var_is_rendered_by_the_server_deployment` (required set **derived** from `config::env`, not restated) | delete the env block from `deployment-server.yaml` | test FAILS, naming the variable |

The end-to-end test builds through the real path rather than asserting on a constant: `from_env` → the controller's own composition function → the real `build_image_build_job` → the Job's **own** shell line evaluated against the Job's **own** container env → the watcher's real `parse_build_metadata`/`classify_ready` → `ImageRepository::mark_ready` → the row read back. Its last assertion is that the stored string equals the Dockerfile's LABEL string.

Also verified: `helm template` with the value set renders `DJINN_IMAGE_LAUNCHER_AUTHORITY_PROTOCOL=resize-v2`; unset renders no such env at all.

## Can a protocol change still silently reuse a cached image?

**No.** A change to the configured protocol changes the environment hash, which changes the image tag, which means the controller's `hash unchanged and ready — skipping` arm cannot be taken and buildkit is pushed to a tag no prior artifact occupies. The one residual path — a build context whose reported protocol disagrees with its Dockerfile — is refused before a Job exists. The remaining assumption is the ordinary one that predates this PR: the launcher binary itself ships from `agent_worker_image`, which is already an input to the same hash.

## Not touched

`server/src/task_run_resize_rollout.rs` and `server/crates/djinn-k8s/src/cutover_preflight.rs` (sibling agent owns the activation entry point). No live cluster was created.

## Notes for review

* `generate_dockerfile` gained a **required** third argument rather than a defaulting overload. A convenience overload is how a future caller silently keeps a configured fleet on `leaf-v1` — the same "merged, green, unreachable" shape this campaign keeps finding. Existing tests keep every assertion and every golden fixture; only the call gains the explicit default.
* `DECLARED_LAUNCHER_PROTOCOL` → `DEFAULT_LAUNCHER_PROTOCOL`: it is no longer what is declared, only what is declared absent a selection.
* `config::test_env::guard` also fixes a pre-existing race between `from_env_honors_zot_retention_vars` and `from_env_invalid_zot_retention_bools_fall_back`, which both write `DJINN_ZOT_RETENTION_ENABLED` on a multi-threaded harness. It surfaced while adding these tests; the lib suite is stable at `--test-threads` 1, 16 and 32.

### DIFF
--- a/deploy/helm/djinn/templates/deployment-server.yaml
+++ b/deploy/helm/djinn/templates/deployment-server.yaml
@@ -387,6 +387,16 @@ spec:
             # later tick as slots free.
             - name: DJINN_IMAGE_MAX_CONCURRENT
               value: {{ .Values.imagePipeline.controller.maxConcurrentBuilds | quote }}
+            {{- with .Values.imagePipeline.controller.launcherAuthorityProtocol }}
+            # Launcher authority protocol newly built catalog images declare in
+            # their own OCI metadata. Rendered only when the operator sets it,
+            # so an unset value is the binary's default (leaf-v1) and no
+            # existing image's environment hash moves. Setting it re-tags and
+            # rebuilds the catalog — a cached leaf-v1 artifact cannot be served
+            # under a resize-v2 declaration.
+            - name: DJINN_IMAGE_LAUNCHER_AUTHORITY_PROTOCOL
+              value: {{ . | quote }}
+            {{- end }}
             # djinn-k8s runtime (task-run + warm-job dispatcher) — same
             # pattern as the image pipeline above: everything is derived
             # from the release fullname + namespace so no hardcoded
--- a/deploy/helm/djinn/tests/kueue-topology-render.sh
+++ b/deploy/helm/djinn/tests/kueue-topology-render.sh
@@ -76,7 +76,9 @@ render_enabled() {
 # EXTRACTED from the Rust that renders the Jobs:
 #
 #   1. Discover the Job renderers: every non-test `.rs` under $CRATES_ROOT that
-#      defines a `pub fn build_*job*(..) -> Job`. Today that is job.rs
+#      defines a `pub fn build_*job*(..) -> Job` — or `-> Result<Job, E>`, since
+#      a renderer that can REFUSE still emits `suspend: true` and still requests
+#      resources on every path that returns one. Today that is job.rs
 #      (task-run), warm_job.rs, scip_job.rs and the image-controller's
 #      build_job.rs — but the LIST is derived, so a fifth renderer joins it
 #      without an edit here.
@@ -400,9 +402,58 @@ _self_test_strip()
 
 
 # A Job renderer is a function that RETURNS a Job. Nothing here names the four.
+#
+# "Returns a Job" includes returning one FALLIBLY. A renderer that can refuse —
+# `build_image_build_job` refuses to render a Job whose build context reports a
+# launcher authority protocol its own Dockerfile does not declare — still emits
+# `suspend: true` and still requests resources on every path that does return,
+# so it is exactly as much of this guard's business as an infallible one. Before
+# `Result` was allowed here the image-build renderer silently dropped out of the
+# discovered set and the floor below caught it, which is the correct direction
+# to fail; matching both shapes is the fix, and `_self_test_job_builder` keeps
+# the fallible shape from regressing back out of view.
 JOB_BUILDER = re.compile(
-    r"pub\s+fn\s+build_[A-Za-z0-9_]*job[A-Za-z0-9_]*\s*\([^{;]*?\)\s*->\s*Job\s*\{", re.S
+    r"pub\s+fn\s+build_[A-Za-z0-9_]*job[A-Za-z0-9_]*\s*\([^{;]*?\)\s*->\s*"
+    # `Job`, or any `Result<Job, E>` — including a path-qualified alias
+    # (`kube::Result<Job>`) and an error type carrying its own generics
+    # (`Result<Job, Foo<Bar>>`), which the lazy body plus the `\{` anchor
+    # resolves to the LAST `>` before the block.
+    r"(?:Job|(?:[A-Za-z0-9_]+\s*::\s*)*Result\s*<\s*Job\s*(?:,[^{;]*?)?>)"
+    r"\s*\{",
+    re.S,
 )
+
+
+def _self_test_job_builder():
+    """The discovery regex, both directions, every run.
+
+    A discovery that quietly matches fewer renderers makes the coverage
+    assertion cover less while still passing — the same silence this guard
+    exists to end — so it is proven before it is used.
+    """
+    for source in [
+        "pub fn build_task_run_job(spec: &Spec) -> Job {",
+        "pub fn build_image_build_job(c: &C) -> Result<Job, DeclarationError> {",
+        "pub fn build_warm_job(\n    a: A,\n    b: B,\n) -> Result<Job, Error> {",
+        "pub fn build_scip_job(a: A) -> kube::Result<Job> {",
+        "pub fn build_x_job(a: A) -> Result<Job, foo::Bar<Baz>> {",
+        "pub fn build_y_job(a: A) -> Result< Job , E > {",
+    ]:
+        assert JOB_BUILDER.search(source), f"renderer not discovered: {source!r}"
+
+    for source in [
+        # Returns something that merely mentions a Job.
+        "pub fn build_job_name(a: A) -> String {",
+        "pub fn build_job_owner_reference(job: &Job) -> Option<OwnerReference> {",
+        # A Result of something else entirely.
+        "pub fn build_a_job(a: A) -> Result<ConfigMap, E> {",
+        # A declaration, not a definition.
+        "pub fn build_task_run_job(spec: &Spec) -> Job;",
+    ]:
+        assert not JOB_BUILDER.search(source), f"falsely discovered: {source!r}"
+
+
+_self_test_job_builder()
 REQUESTS_BLOCK = re.compile(
     r"requests\s*:\s*Some\(\s*BTreeMap::from\(\s*\[(?P<body>[^\[\]]*)\]\s*\)\s*\)", re.S
 )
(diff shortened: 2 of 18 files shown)


fernandobandeira:
## Guard fix pushed (d1bbfab93)

`Local Dev Contracts / Test Helm chart contracts` was the one real failure; the other reds were fail-fast cascade.

**Cause was mine and the guard was right to fire.** `deploy/helm/djinn/tests/kueue-topology-render.sh` (zxdv, #2855) discovers Job renderers by matching `pub fn build_*job*(..) -> Job`, then derives the resource keys they request and asserts the ClusterQueue's `coveredResources` is a superset. `build_image_build_job` now returns `Result<Job, DeclarationError>`, so it stopped matching and the image-build renderer dropped out of the discovered set — 4 became 3, and the "at least four" floor caught it. Failing loudly there is exactly the drift that guard exists to prevent: a discovery that quietly covers fewer renderers leaves the coverage assertion checking less while still passing.

**Fixed the discovery pattern, not the return type.** A renderer that can *refuse* still emits `suspend: true` and still requests resources on every path that returns a Job, so it is exactly as much of the guard's business as an infallible one. The pattern now matches `-> Job` and `-> Result<Job, E>`, including a path-qualified alias (`kube::Result<Job>`) and an error type carrying its own generics (`Result<Job, Foo<Bar>>`).

Explicitly **not** done: no revert of the `Result` return (the refusal it enables — a Job carrying a Dockerfile label and a catalog row that disagree — is worth more than the guard's convenience); no hardcoded renderer paths; no lowered floor. Both would pass vacuously the day a fifth renderer appears.

**Regression proof in the other direction:** `_self_test_job_builder` runs on every invocation alongside the existing `_self_test_strip`, asserting both directions — the fallible signature is discovered, and `build_job_name(..) -> String`, `build_job_owner_reference(..) -> Option<OwnerReference>`, `-> Result<ConfigMap, E>` and a bare declaration are not.

Verified locally:
- Reverting the pattern to `-> Job` only: `AssertionError: renderer not discovered: 'pub fn build_image_build_job(c: &C) -> Result<Job, DeclarationError> {'`
- With the fix: all four renderers rediscovered (`build_job.rs`, `job.rs`, `scip_job.rs`, `warm_job.rs`), `=== All Kueue topology Helm render tests passed ===`
- Full `scripts/test-helm-chart-contracts.sh`: 16/18 pass; the two failures are `FAIL: vector is required to execute VRL fixtures` (missing local binary, present in CI), unrelated to this branch.

`kueue-topology-render.sh` is zxdv's file and outside this PR's original set — noted here as expected collateral of the `Result` return.

fernandobandeira:
## Capability boundary fixed without an exemption (5aba7df50)

`Server Guards / Self-test capability boundary detectors` was the earliest real failure: `declared_protocol_tests.rs` named `k8s_openapi` from inside `djinn-image-controller`. **No allowlist entry added, no detector widened, nothing `#[allow]`ed past** — the image-controller's k8s exceptions are an inventory being shrunk under epic/fztz, not a list to append to for a test's convenience.

Took option 3 (assert without the k8s type) plus option 2 for the one assertion that genuinely needs the rendered object:

- **`render_builder_script` extracted** out of `build_image_build_job`. The builder's shell is now reachable as a `String`, so the end-to-end test reads the **real** script text instead of restating its echo line.
- **One spelling for the seam.** `LAUNCHER_PROTOCOL_JOB_ENV` is a single constant read by both the container env `build_image_build_job` sets and the `echo` the script expands; `watcher::PROTOCOL_SENTINEL` is now `pub(crate)` so the emitter and the parser of the sentinel share one spelling. A rename on one side is no longer expressible.
- **`declared_protocol_tests.rs` names no Kubernetes type.** The chain is: config → `render_build_context` → Dockerfile LABEL/ENV → `verify_declaration` → the script's own echo, expanded → `parse_build_metadata` → `classify_ready` → `mark_ready` → the row, compared against the Dockerfile's own LABEL string.
- **`build_job.rs` — already inside the inventory — gains `the_build_container_env_carries_the_configured_declaration`:** for every protocol, the rendered Job's container env carries the configured declaration and agrees with the artifact's LABEL, against a tag that lies in the opposite direction.

### verify_declaration's teeth are intact, re-proven by mutation

| Mutation | Result |
|---|---|
| drop `verify_declaration()?` from `build_image_build_job` | `a_build_whose_label_and_catalog_row_would_disagree_is_refused` **FAILS** — the mismatched label/row builds instead of being refused |
| render the Job env from a literal instead of the verified declaration | `the_build_container_env_carries_the_configured_declaration` **FAILS** |
| give the script its own env-var spelling | **3 tests FAIL**, including the end-to-end chain — the Pod would print an unexpanded `${…}` the watcher refuses. The simplest form of this mutation doesn't even compile (unused named format argument) |

`scripts/test-capability-boundaries.sh`: **77 passed, 0 failed**, including the exact failing case `baseline-T3 k8s detector full tree exits 0`. Allowlist entries unchanged: `git=14 http=0 k8s=7`.

Also re-ran green locally: `kueue-topology-render.sh`, `check-resize-reachability.sh`, `check-resize-cutover-retention.sh`, `check-resize-authorization-boundary.sh`, `cargo fmt --check`, `clippy -D warnings`, and the full image-builder + image-controller suites.

PR #3093, opened 2026-08-07
nafu W3b: lane executors, provider seam, and the ci_evidence_routing gate

fernandobandeira:
**Wave 3b of proposal [nafu](https://code.djinnai.io/proposals/019fccac-e07d-7493-a1bc-18996e7d439a)** — the two lane executors, the provider seam, and the real runtime gate.

**Stacked on #3092 (W3).** Merge that first. W5 stacks on this.

## What lands

- **`CiRouteProvider` trait** over `rerun_failed_jobs`, `enable_auto_merge`, `list_check_runs_for_ref`, and `get_check_run_annotations`, following the existing `PrCleanupGitHub` pattern exactly. The fourth method is the **producer for `LogApiError`**, which had none anywhere in the tree. No provider behaviour changed.
- **Both lane executors, atomically** — PR-head reuses `PrDraftCiAction::RetriggerAndHold` + `rerun_failed_jobs`; merge-group calls `enable_auto_merge` and replaces the immediate `handle_queue_failure` only for Tier 1.
- **`ci_evidence_routing`** — a real three-state gate (`DJINN_CI_EVIDENCE_ROUTING`, default `disabled_clean`). Previously the feature was off only by *absence*: no flag existed anywhere.
- **All ten spec-named fixtures** now exist under their spec names, each discard counting four negatives against real rows — provider mutations from the fake, leases and route rows from the route table, `tasks.status` + `activity_log` for board mutation, `task_attempts` for worker dispatch.

## Two live production bugs fixed

- **`pr_commands.rs` matched only `Some("failure")`** while the shared predicate matches `failure|timed_out|cancelled`. A merge-group run that **timed out** got no snapshot row and no strike count, so it **never parked** — an infinite blind-reopen loop, with zero tests on the live path. Now uses `correlate_merge_group_run`, which is ambiguity-aware rather than `find`-ing over unpinned ordering.
- **`list_check_runs_for_ref` forged `Complete`** — single-page, no pagination, `#[serde(default)]`. A ref with >100 check runs truncated *and claimed completeness*, defeating W3's contract.

## Audited, then fixed

The audit found six blockers, all closed and each re-verified by mutation. The most serious: **`from_env`'s default had zero coverage** — changing `unwrap_or_default()` to `unwrap_or(Enabled)` turned the feature on fleet-wide with 2148 tests green. Also closed: a `CompleteEmpty` decision produced and never consumed (a gate-on wedge stranding `pr_review` tasks), untested lane call sites, and a mislabeled owner-handoff fixture that never aged its row.

One knowingly-unsealed edge, documented rather than claimed: `CiEvidenceIdentity::run_id`/`lane` are public on the `djinn-db` type, so the action scope is bypassable by a determined caller. The structural claim was **removed** from the doc comment rather than asserted falsely.


fernandobandeira:
CI run cancelled deliberately — this PR stacks on #3092 and must merge after it. Holding to avoid runner starvation (six concurrent quality-gate runs against a ~60-shard cap). Will rebase and re-run once #3092 lands.

PR #3095, opened 2026-08-07
nafu W5: supervisor apply guard, reporting, rollback report, and rev-58 reconciliation

fernandobandeira:
**Wave 5 — the final wave of proposal [nafu](https://code.djinnai.io/proposals/019fccac-e07d-7493-a1bc-18996e7d439a)**, reconciled to the settled spec at **revision 58**.

**Stacked on #3092 (W3) and #3093 (W3b).** Merge those first. Contains W4's work, already merged as #3080.

## The supervisor apply guard — the reason this wave exists

Before this, the halves were split the wrong way: `stage.rs::lead_stage_outcome_routed` shipped **live** while `board_effect`, `durable_resolution`, and `CiGuardOutcome` were dead with zero callers, and W1's `resolve_tier2_lease` had zero production callers workspace-wide. The supervisor applied `StageOutcome::LeadReopen` immediately with **nothing in between**.

Wiring it exposed a further gap neither prior wave reported: **the executor opened the Tier-2 lease and stopped.** Nothing dispatched Lead and nothing wrote the `ci_route` block, so `from_arbiter_directive` had no producer anywhere and answered `None` on every production call — a Tier-2 route left the task with the legacy path suppressed and no remedy in its place.

The chain is now: `CiTier2Handoff` → `tier2_dispatch` writes the arbitration row and escalates (board transition **last**, so a failed directive write cannot dispatch a Lead with no block) → `read_arbiter_directive` → `apply_under_guard` calls `resolve_tier2_lease` with a re-read PR head → the supervisor applies **only if the CAS held**. A failed guard yields `LeadRouteSuperseded`, which needed its own variant: `Failed` on a Lead role feeds `record_arbiter_session_termination` and would **park a task whose only sin was a head that moved twice**.

## Reconciliation to rev 58

The tribunal amended the incomplete-evidence contract after this code was written. The discriminator is now **recoverability, not completeness**:

- **Recoverable** (failed page, short read, correlation not yet observable) → bounded hold, escalating at the specified consecutive-poll limit.
- **Irrecoverable** (`MAX_PAGES` truncation, `RunAttributionUnavailable`, ambiguous correlation, the four timestamp reasons, escalated holds) → **one diagnose-only Tier 2** under a **run-absent identity**.

Plus the ordering contract the last objection forced: a short transaction assigns each logical poll UUID one monotonic sequence **before** provider enumeration; a second applies only when the sequence exceeds the retained high-watermark; **no transaction spans the provider call**; complete evidence resets the count without deleting the row; retries of one poll reuse one sequence.

Also: the `run_id: 0` sentinel is replaced by nullable `run_id` under `CHECK (run_id IS NULL OR run_id > 0)` with `UNIQUE NULLS NOT DISTINCT`; budget monotonicity moves from Rust convention to a `BEFORE UPDATE` trigger; `attach_lead_session` becomes append-only; `parks_with_cited_cause` reads the column it is named after.

## Reporting and rollback

Every adjudication count reads `COALESCE(tier2_resolution, terminal_outcome)` — reading `terminal_outcome` alone silently drops every reopen that followed a provider failure. A durable `lead_rejection` column separates *"Lead diagnosed"* from *"Lead never answered"*, which were otherwise byte-identical. The rollback report pairs four repository counts with the two only the coordinator can attest, under a DB CHECK recomputing `permits_rollback` so a green report cannot be persisted over a live `calling` row.

## Verification

AC12, all five commands, exit 0 with positive counts: **119 / 53 / 11 / 47 / 8**. Clippy clean across five crates; `phase1.json` byte-identical.

Mutation-proven: the ordering contract, the `RunAttributionUnavailable` escalation, the run-absent identity, and the budget trigger.

**Migration 195** — 194 was taken on main by #3056; 193 is byte-identical to main and hash-pinned. Note the immutability guard cannot catch an edit to a migration added on a squash-merged ancestor: it diffs three-dot against the merge base, so such a file classifies as `Added` and is excluded by `--diff-filter=MRD`.


fernandobandeira:
CI run cancelled deliberately — this PR stacks on #3092 and #3093 and must merge last. Holding to avoid runner starvation. Will rebase and re-run once the parents land.

PR #3178, opened 2026-08-13
fix(k8s): clone the warm workspace from the local mirror, not GitHub

thales-altpay:
## Problem

The warm Job's shell wrapper resolves `remote.origin.url` out of the mirror config and clones from GitHub:

```sh
UPSTREAM_URL="$(git -C "$MIRROR" config remote.origin.url)"
git clone --depth 1000 --single-branch "$UPSTREAM_URL" "$PROJECT_ROOT"
```

So **every warm Pod re-downloads the whole repository**, even though a full bare mirror is already mounted read-only at `/mirror`.

The comment justifying this says a `--local --shared` clone of the mirror would give a partial clone whose checkout fails on missing blobs. **That is no longer true.** `MirrorManager::create_mirror` stopped passing `--filter=blob:none`, and `ensure_full_mirror` promotes pre-existing blobless mirrors on server boot. The task-run path already depends on this via `MirrorManager::clone_ephemeral` — only the warm path still goes to the network.

## Why it matters

On a live staging cluster I attributed VPC flow-log egress by Karpenter node pool. Warm Pods cloning from GitHub were the **single largest source of egress from the cluster**, well ahead of CI. It's pure waste: the bytes are already on the PVC.

## Change

Clone from the mirror with `--local --shared`. The workspace borrows the mirror's object db via `.git/objects/info/alternates`, so nothing crosses the network and the workspace is near-free to create.

This also removes the need for the shallow-history heuristic. `--depth 1000` existed because the coupling-index phase walks `cursor..HEAD` and needs the saved cursor reachable, with `coupling_index::try_fetch_cursor` falling back to `git fetch --unshallow` for older cursors. An alternates clone exposes the mirror's **complete** history, so no cursor can fall outside the window and that fallback stops being reachable from here.

## Verification

Run inside a **running warm Pod**, against a real mirror, as the Pod's own uid:

```
git clone --local --shared /mirror/<project>.git /tmp/lc
  -> 659 files checked out
  -> workspace .git 224K            (vs 44M for the GitHub clone)
  -> rev-list --count HEAD = 1100   (full history, vs --depth 1000)
  -> alternates -> /mirror/<project>.git/objects
```

All live mirrors confirmed full — `extensions.partialClone`, `remote.origin.partialclonefilter` and `remote.origin.promisor` all unset.

Tests updated to pin the new contract: the clone must be `--local --shared` off the mounted mirror, and must not resolve the upstream URL, set a depth, or restrict to a single branch.

## Notes for review

- The mirror stays mounted **read-only**; `--shared` only ever reads it.
- The existing `GIT_CONFIG_SYSTEM` / `safe.directory` setup is unchanged and still required, since git only honours `safe.directory` from protected file scope in the inner child of `git clone --local`.
- The workspace does depend on the borrowed objects surviving for the life of the Pod. That holds because the mirror fetcher only adds refs, and a warm Pod is short-lived relative to any mirror maintenance — but flagging it explicitly in case you'd rather trade the alternates for a plain `--local` copy.
- **Please let CI be the judge. The runtime behaviour above was verified directly in-cluster.

### DIFF
--- a/server/crates/djinn-k8s/src/scip_job.rs
+++ b/server/crates/djinn-k8s/src/scip_job.rs
@@ -216,22 +216,43 @@ pub fn build_scip_index_job(
     let mirror_path = format!("{MIRROR_MOUNT_DIR}/{project_id}.git");
 
     // Deliberately the same clone preamble as the warm Job (umask, the
-    // GIT_CONFIG_SYSTEM safe.directory file, --depth 1000 --single-branch,
+    // GIT_CONFIG_SYSTEM safe.directory file, the `--local --shared` clone of
+    // the mounted mirror followed by an explicit detached checkout, the
     // lockfile-gated JS install). The two Pods index the same tree; a
-    // divergence here is a divergence in what gets indexed.
+    // divergence here is a divergence in what gets indexed — and the two
+    // coordinate through `WARM_SCIP_CLAIM_WAIT_ENV` on exactly that
+    // assumption, with the SCIP artifacts landing in a cache keyed by
+    // revision.
+    //
+    // The mirror clone is what makes "the same tree" true rather than
+    // hopeful. This used to resolve `remote.origin.url` and clone from GitHub,
+    // so the two Pods raced the *remote's* head independently; now both borrow
+    // the same on-PVC object db and both check out a SHA that was read out of
+    // `refs/heads/main` of that same mirror. `revision` here is the SHA this
+    // Job was dispatched for (`MirrorHead::revision`, and the value stamped
+    // into `ANNOTATION_SCIP_REVISION` that the scheduler reads back), so the
+    // tree indexed is by construction the tree the cache entry is named after.
+    //
+    // The explicit `checkout --detach` is not decoration: a bare mirror's HEAD
+    // symref is written once at `git clone --bare` time and never refreshed,
+    // so after an upstream default-branch rename `git clone --local` warns,
+    // EXITS 0, and leaves a directory containing only `.git`. See the long
+    // note in `warm_job::build_warm_job` for the full mechanism, including why
+    // `--shared`'s borrowed objects outlive the Pod.
     let cmd = format!(
         r#"set -euo pipefail
 umask 0002
 export GIT_CONFIG_SYSTEM={WORKSPACE_MOUNT_DIR}/.djinn-gitconfig
 unset GIT_CONFIG_NOSYSTEM
 printf '[safe]\n\tdirectory = *\n' > "$GIT_CONFIG_SYSTEM"
-UPSTREAM_URL="$(git -C "{mirror_path}" config remote.origin.url)"
-git clone --depth 1000 --single-branch "$UPSTREAM_URL" "{project_root}"
+git clone --local --shared "{mirror_path}" "{project_root}"
+git -C "{project_root}" checkout --detach {revision} --
 cd "{project_root}"{js_install}
 exec {bin} scip-index "{project_id}"
 "#,
         mirror_path = mirror_path,
         project_root = project_root,
+        revision = revision,
         bin = WARM_COMMAND_BIN,
         project_id = project_id,
         js_install = crate::js_install::js_install_preamble(&project_root, js_workspace_roots),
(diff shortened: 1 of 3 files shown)


fernandobandeira:
Thanks for this — the direction is right and the reasoning in the comments is unusually careful. The CLA check was failing only because the commit author email (`thales@Thaless-MacBook-Pro.local`) didn't resolve to a GitHub account; I re-pointed it at your `users.noreply.github.com` address, so that's green now and your signature from June was always on file.

A detailed review turned up one blocking defect and one loose end before this can land.

**Blocking — the new clone can exit 0 and produce an empty workspace.** Reproduced locally:

```
mirror.git: HEAD -> refs/heads/main, but only refs/heads/trunk exists
$ git clone --local --shared mirror.git ws
warning: remote HEAD refers to nonexistent ref, unable to checkout
done.
EXIT=0
$ ls -A ws
.git            # that's the whole working tree
```

Mirrors get `HEAD` set once at `git clone --bare` time (`mirror.rs:176-186`) and nothing ever updates it afterwards — there's no `remote set-head`/`symbolic-ref` write anywhere in the repo — while `fetch_mirror` runs with `--prune`. So an upstream default-branch rename (`master`→`main`) leaves the mirror's HEAD permanently dangling. The warm Pod then clones successfully, `cd`s into a directory containing only `.git`, skips the lockfile-gated JS install, and `exec`s `warm-graph` against an empty tree — silently, for every warm on that project, with no error to grep for. The old `git clone --single-branch "$UPSTREAM_URL"` was immune because it resolved the remote's live HEAD over the wire. `set -euo pipefail` doesn't help; git really does exit 0.

Notably this would make the warm Job the only `--local --shared` caller in the codebase that trusts the mirror's HEAD: `clone_ephemeral` passes `--branch` (`mirror.rs:452`), `ensure_workspace` passes `--branch` (`workspace_store.rs:234`), and `clone_ephemeral_at_ref` follows up with `git checkout --detach <sha>` (`mirror.rs:534`).

**Fix:** make the checkout explicit. `warm_generation` is already a parameter of `build_warm_job` and is already the mirror's `refs/heads/main` SHA, so following the clone with `git -C "$PROJECT_ROOT" checkout --detach {warm_generation}` (or cloning `--branch main`) costs nothing and fixes two things at once — it turns the silent-empty case into a loud failure, and it guarantees the Pod indexes exactly the revision `djinn.app/graph-revision` claims. Right now the clone checks out whatever the frozen HEAD symref names, while `discover_mirror_main_tip` hardcodes `refs/heads/main` (`graph_warmer.rs:1814`); nothing guarantees those agree.

Please **don't** switch to `--local` without `--shared`, or add a post-clone `git repack -a -d` — djinn's pack is 2.67 GiB, the mirror files are uid 10001 while the Pod runs as uid 1001, so `fs.protected_hardlinks=1` makes git silently fall back to byte copies. That's a 2.67 GiB copy into emptyDir on every warm. `--local --shared` is the right call.

**Loose end — `scip_job.rs` still has the old command** (`scip_job.rs:228-229`), and its own comment says *"Deliberately the same clone preamble as the warm Job… a divergence here is a divergence in what gets indexed."* After this lands the warm Pod indexes the mirror's head while the SCIP Pod indexes GitHub's head, and the two coordinate through `WARM_SCIP_CLAIM_WAIT_ENV` on the assumption they share a tree. Either apply the same change there or update that comment to record why they now differ.

**One comment to correct:** *"the mirror fetcher only ever adds refs"* (`warm_job.rs:192-195`) isn't accurate — `fetch_mirror` runs `--prune`, `delete_branch` removes task refs, and `git_maintenance.rs` runs a daily leader-only `git gc --prune=2.weeks.ago` against that same mirror. The conclusion still holds, but for a subtler reason worth writing down: `git gc` freshens the pack mtime even on a no-op repack, and the prune window is measured from pack mtime, so with a daily cadence newly-unreachable objects do get a real grace period. Stated as-is, the comment will get quoted back as license the next time someone widens this.

Worth adding: cloning from the mirror is a **correctness** win you didn't claim. The old code indexed GitHub's live head while the Job was annotated with the mirror revision it was dispatched for; this makes them agree. That's a stronger argument than the egress one.

PR #3181, opened 2026-08-17
fix(breaker): give the model hard-disable a half-open probe instead of a permanent latch

fernandobandeira:
## The outage

The autonomous build loop was dead for four days, 2026-08-12 → 2026-08-16.

`openai/gpt-5.6-terra` was the **only** candidate in the user's `implement` lane. A transient provider incident on 08-12 pushed it past the trip-rate ceiling (`TRIP_RATE_CEILING` = 8 trips within `TRIP_RATE_WINDOW` = 6h) and it latched `hard_disabled: true`. Every worker dispatch after that went `breaker_open` → `failover_chain_exhausted` → cooldown, forever.

On 08-16 the bucket read:

```
hard_disabled: true, auto_disabled: true, disable_ttl_trips: 8,
cooldown_seconds_remaining: null, trips_in_window: 0
```

**`trips_in_window: 0`.** Every trip that justified the latch had already aged out of the rolling 6-hour window, yet the latch was still set. A manual reset produced 10/10 successes immediately, on top of 1753 completions in the trailing 30-day window.

The defect: **a rate-based trigger with a permanent, never-re-evaluated consequence.** `ModelState::is_available` began with an unconditional `if self.hard_disabled { return false; }`, and `register_trip` explicitly cleared `cooldown_until` so nothing else could ever clear it.

## The tension

The permanence was **not** gratuitous, and this PR does not delete it. A model that trips 8×/6h "is broken, and continuing to auto-re-enable it just keeps feeding it priority tasks it cannot complete." There is a documented prior incident of exactly that flapping — disable 5 min → re-enable → grab a slot → crash in <10s → repeat, with **one production task losing 5.75h to 8 consecutive 429 crashes**.

So the requirement is narrow: **recover from a decayed transient without reintroducing the flap.** Two rounds of adversarial review were spent finding places where the second half of that sentence quietly failed.

## The design

| | |
|---|---|
| **Quarantine base** | 6h, anchored at the bucket's most recent trip |
| **Admission** | exactly **one** probe dispatch per quarantine period |
| **Failed probe** | re-latch + double the quarantine (6h → 12h → 24h → 48h → 96h → 7d ceiling) |
| **Successful probe** | release to **probation**, not to a clean slate |
| **Probation** | win 5 of at most 6 sessions, else re-quarantine one rung higher |
| **Restart** | absolute wall-clock deadline — neither wiped nor extended |
| **Human `enable`/`reset`** | immediate and authoritative at any point |

### Why 6 hours

Three bounds, two enforced at compile time by `const _: () = assert!(...)`:

1. **It must exceed `MAX_COOLDOWN` (4h)**, or the hard-disable is indistinguishable from the escalating ladder it backstops.
2. **It equals `TRIP_RATE_WINDOW`, and that equality is load-bearing.** Because the deadline is anchored at the most recent trip, by the time the probe fires every trip that justified the latch has been pruned — `trips_in_window` is provably `0` at probe time. The probe asks "is this model healthy **now**" against a window that no longer contains any of the evidence for the latch. That is exactly the state the incident bucket sat in, ignored, for four days.
3. **Exposure is negligible**: one task per quarantine period, escalating to one per week.

### Probation: why a probe success is not a pardon

A probe success is real evidence, so the lane reopens at **full throughput** — that is what fixes the outage. But it is evidence from one session about a bucket that reached eight trips in six hours. The first version treated it as a full pardon (clearing the tier, trip window and cooldown ladder), which let an intermittently-broken model escape the escalation ladder entirely.

Probation preserves the tier and runs a **counted, bounded trial**: win `HARD_DISABLE_PROBATION_SUCCESSES` (5) of at most `HARD_DISABLE_PROBATION_TRIAL_SESSIONS` (6). Successes and failures are counted independently and neither resets the other, so **failure clustering is irrelevant** — the flaw that survived the first review. The trial resolves three ways, all inside the budget:

* 5 wins → history forgiven;
* `probation_failure_ceiling()` (derived: 2) losses → re-quarantined one rung higher;
* budget spent without either → re-quarantined.

The third clause is charged in `note_dispatch_accepted` — on **exposure**, not outcome — so the bound holds even when sessions never report anything. Exposure per released quarantine is at most `1 + 6` sessions **by construction, against any input**.

A genuinely recovered model (the outage's went 10-for-10) wins its trial with full throughput throughout. A one-success-in-six model must win ≥5 of 6: `C(6,5)·(1/6)^5·(5/6) + (1/6)^6 ≈ 0.00066`, about 1 in 1500.

### Restart: an absolute wall-clock deadline

`Instant`s are meaningless across a process boundary. Re-anchoring at `now + quarantine_for_tier(tier)` looked conservative and was not — it charged a full fresh quarantine every boot. Measured probes admitted in 30 days by deploy cadence: **1h→0, 4h→0, 5h→0, 6h→1, 8h→1**, against a real cadence of median 6.08h with 12 of 24 gaps under 6h. The escalation ladder made starvation *more* certain, not less. That is the original outage through a different door.

The quarantine now carries `hard_disable_probe_at_unix` and `restore_all` restores the remainder from it: downtime counts exactly as uptime would. The remainder is clamped to the tier's quarantine, so a corrupt or skewed snapshot can shorten a quarantine but never manufacture a longer one.

### Durability

`persist_model_health_state` previously had three callers, all human admin actions — no periodic persist, none on shutdown. Every "persisted" property was only as durable as the last time an operator ran a command. `AppState::run_model_health_persist_loop` now writes on a 5-minute cadence and once more on cancellation.

### `enable` vs `reset`

Both remain authoritative and immediate. `enable` additionally clears the failure **streaks** while preserving lifetime totals: it used to leave `breaker_eligible_consecutive_failures` at or above `CIRCUIT_BREAKER_THRESHOLD`, so the next failure re-tripped on the spot and every `enable` had to be chased with a `reset`. `disable_ttl_trips` is deliberately kept — the operator vouched the model is usable now, not that its history never happened.

### Operator surface

`model_health` now reports `hard_disable_probe_seconds_remaining` (`Some(0)` = probe admissible now, `None` = not quarantined), `hard_disable_on_probation` and `hard_disable_probe_tier`, and the `hard_disabled` doc no longer claims "no auto-expiry until a human re-enables". This is the surface that was actually used to diagnose the outage, where `hard_disabled: true, trips_in_window: 0, cooldown: null` gave no way to tell a permanent state from a wait.

## The exposure table

30 simulated days, one-minute polling, driving the real API in the real order (`is_available` → `note_dispatch_accepted` → outcome). Asserted in `probation_bounds_exposure_across_every_failure_pattern_not_just_one`:

| pattern | this PR | after round-2 fixes (clustering hole) | `origin/main` |
|---|---|---|---|
| one success in 2 | **32** · quarantined · tier 5 | 42840 · never re-quarantined | 0 · permanent latch |
| one success in 3 | **24** · quarantined · tier 5 | 42840 · never re-quarantined | 0 |
| one success in 4 | **16** · quarantined · tier 5 | 32 | 0 |
| one success in 6 | **12** · quarantined · tier 5 | 17 | 0 |
| F,F,S repeating | **20** · quarantined · tier 5 | 40682 · never re-quarantined | 0 |
| S,S,F repeating | **48** · quarantined · tier 5 | 42840 · never re-quarantined | 0 |
| alternating S,F | **32** · quarantined · tier 5 | 42840 · never re-quarantined | 0 |
| all failures | **8** · quarantined · tier 5 | 8 | 0 |

Every row asserts four properties, not one: bounded exposure, ends quarantined, escalated past tier 0, and **still probed at all** (a zero would be the 4-day outage, not a fix). The structural worst case is `8 cycles × (1 probe + 6 trial sessions) = 56`; the ceiling is 80.

The previous round asserted a single hardcoded `success_every = 6`, passed at 17, and `success_every = 3` went through the identical assertion shape at 42840. That is why this is a table.

## Which tests prove what

| test | defect |
|---|---|
| `quarantine_admits_exactly_one_probe_and_a_clean_probe_recovers_the_model` | the incident; fails on `main` |
| `probation_bounds_exposure_across_every_failure_pattern_not_just_one` | N1 clustering |
| `probation_re_quarantines_on_counted_failures_regardless_of_clustering` | N1 mechanism |
| `a_probation_trial_is_bounded_by_dispatches_even_with_no_outcomes_at_all` | N1 unbounded window |
| `probation_tightens_the_throttle_ladder_too` | N4 |
| `stray_successes_on_a_quarantined_bucket_cannot_pre_charge_the_probation_trial` | N2 |
| `repeated_restarts_inside_the_quarantine_cannot_starve_the_probe` | N3 starvation |
| `a_restart_preserves_the_quarantine_deadline_in_both_directions` | N3 both directions + skew clamp |
| `a_success_without_an_admitted_probe_cannot_release_a_quarantine` | D2 |
| `health_blind_failures_cannot_postpone_the_probe_forever` | D3 |
| `a_transient_upstream_fault_re_quarantines_without_escalating` | D4 |
| `model_health_reports_the_next_probe_deadline_to_operators` | D5 |

### Surgical reverts

| reverted | tests that fail | measured |
|---|---|---|
| N1 counted trial → trip-based | 5 | table reads 42840 / 42840 / 32 / 17 / 40682 — matching the review exactly |
| N2 ledger reset | 4 | full pardon from four stray successes + one probe |
| N3 wall-clock deadline | 2 | 0 probes vs 8 under an hourly deploy |
| D1 probation release | 3 | 42840, never re-latches |
| D2 `probe_outstanding` on success | 1 | latch released with 0 probes admitted |
| D3 `probe_outstanding` on failure | 13 | 0 probes vs 6 for the quiet baseline |
| `is_available` half-open branch | 8 | permanent latch restored |

Restored: **68/68 pass**.

## Test results

| Command | Result |
|---|---|
| `-p djinn-provider --lib catalog::health` | **68 passed / 0 failed** |
| `-p djinn-provider --lib` | **630 / 0** |
| `-p djinn-control-plane` | **1022 / 0** across 12 targets |
| `-p djinn-server --lib` | **665 / 0** (2 ignored) |
| `-p djinn-coordinator --features test-support --lib` | 2308 / 1 — load-sensitive, passes in isolation, identical on `origin/main` under the same invocation |
| `cargo clippy --workspace --all-targets --features qdrant --keep-going` | clean |
| `cargo fmt --all`, `git diff --check` | clean |
| `make tool-goldens-check` | up to date |
| raw-SQL boundary + self-test | OK / 59 passed 0 failed |

Private PostgreSQL on port 55219, torn down. All work inside the agent worktree.

## Not verified

* No staging/production run.
* The persist loop is wired and compiles, but its 5-minute cadence and shutdown write are not covered by an integration test — `djinn-server` has no harness that boots `AppState` and advances time.
* `HARD_DISABLE_PROBATION_SUCCESSES = 5` and `HARD_DISABLE_PROBATION_TRIAL_SESSIONS = 6` are reasoned, not tuned against production data. The binding constraint is the trial length, not either constant alone.
* `evidence_dispatch_recovery` still dispatches without consulting the breaker at all. It can no longer corrupt the quarantine (D3), but that gap is pre-existing, in another module, and deserves its own fix.

### DIFF
--- a/server/crates/djinn-control-plane/src/tools/provider_tools.rs
+++ b/server/crates/djinn-control-plane/src/tools/provider_tools.rs
@@ -184,12 +184,35 @@ pub struct ModelHealthOutput {
     pub disable_ttl_trips: u32,
     #[schemars(with = "Option<i64>")]
     pub cooldown_seconds_remaining: Option<u64>,
-    /// Hard-disabled by the trip-rate ceiling: held unavailable with no
-    /// auto-expiry until a human re-enables it via `model_health(action=enable)`.
-    /// When true, `cooldown_seconds_remaining` is null (there is no auto-expiry).
+    /// Hard-disabled (quarantined) by the trip-rate ceiling: held unavailable
+    /// except for a single half-open probe dispatch admitted once per
+    /// quarantine period. When true, `cooldown_seconds_remaining` is null (the
+    /// ordinary cooldown ladder does not own this bucket) and
+    /// `hard_disable_probe_seconds_remaining` says when the next probe is due.
+    /// A human can still release it immediately with
+    /// `model_health(action=enable)`.
     pub hard_disabled: bool,
+    /// Seconds until the next half-open probe dispatch is admitted; `0` means a
+    /// probe is admissible right now, and `null` means the bucket is not
+    /// quarantined. Read this before reaching for `enable`: a `hard_disabled`
+    /// bucket recovers on its own if its next probe succeeds. Its absence is
+    /// what made the 2026-08-12 → 08-16 outage unreadable — `hard_disabled:
+    /// true, trips_in_window: 0, cooldown: null` gave an operator no way to tell
+    /// a permanent state from a wait.
+    #[schemars(with = "Option<i64>")]
+    pub hard_disable_probe_seconds_remaining: Option<u64>,
+    /// The quarantine was released by a successful probe, but the bucket has not
+    /// yet produced enough clean sessions to have its history forgiven: it is
+    /// dispatchable at full throughput, and a single further breaker trip
+    /// re-quarantines it at the next (longer) tier.
+    pub hard_disable_on_probation: bool,
+    /// Escalation tier of the quarantine ladder: `0` = 6h, doubling per failed
+    /// probe to a 7-day ceiling. Survives restarts.
+    #[schemars(with = "i64")]
+    pub hard_disable_probe_tier: u32,
     /// Number of breaker trips inside the rolling trip-rate window (6h). When it
-    /// reaches the ceiling (8) the bucket hard-disables.
+    /// reaches the ceiling (8) the bucket hard-disables — or `1` while
+    /// `hard_disable_on_probation` is set.
     #[schemars(with = "i64")]
     pub trips_in_window: u32,
 }
@@ -206,6 +229,9 @@ impl From<ModelHealth> for ModelHealthOutput {
             disable_ttl_trips: value.disable_ttl_trips,
             cooldown_seconds_remaining: value.cooldown_seconds_remaining,
             hard_disabled: value.hard_disabled,
+            hard_disable_probe_seconds_remaining: value.hard_disable_probe_seconds_remaining,
+            hard_disable_on_probation: value.hard_disable_on_probation,
+            hard_disable_probe_tier: value.hard_disable_probe_tier,
             trips_in_window: value.trips_in_window,
         }
     }
--- a/server/crates/djinn-coordinator/src/dispatch/task_dispatch.rs
+++ b/server/crates/djinn-coordinator/src/dispatch/task_dispatch.rs
@@ -1090,6 +1090,15 @@ impl CoordinatorActor {
 
             match dispatch_fn(&self.pool, model_id).await {
                 Ok(()) => {
+                    // A dispatch was actually accepted onto this model. If the
+                    // bucket is hard-disabled this is the ONE half-open probe
+                    // its quarantine admitted, and consuming it here — at the
+                    // single point where a real task is exposed, not at the
+                    // `is_available` predicate above, which is a pure read
+                    // consulted several times per decision — re-arms the
+                    // quarantine before the probe's verdict is known. No-op for
+                    // a healthy bucket.
+                    self.health.note_dispatch_accepted(scope, model_id);
                     tracing::Span::current().record("outcome", "ok");
                     tracing::info!(outcome = "ok", model_id = %model_id, label);
                     super::lane_resolution_log::emit_failover_candidate_accepted(
--- a/server/crates/djinn-provider/src/catalog/health.rs
+++ b/server/crates/djinn-provider/src/catalog/health.rs
@@ -169,6 +169,48 @@ const _: () = assert!(
     "HARD_DISABLE_MAX_PROBE_TIER must be large enough to actually reach the ceiling"
 );
 
+/// Trips required to re-quarantine a bucket that is on **probation** — i.e. one
+/// whose latch was released by a successful half-open probe but which has not
+/// yet re-proven itself (see [`HARD_DISABLE_PROBATION_SUCCESSES`]).
+///
+/// A probation bucket has already demonstrated it can reach
+/// [`TRIP_RATE_CEILING`], and it was handed the lane back on the strength of a
+/// SINGLE session. Making it earn eight fresh trips before it can be
+/// re-quarantined is what turns a lucky probe into a flap: the first version of
+/// this change cleared `trip_times`, `disable_ttl_trips` and the quarantine tier
+/// on a probe success, leaving the bucket byte-identical to a never-tripped
+/// model. Measured against the `xiaomi-token-plan-sgp/mimo-v2.5-pro` profile
+/// (78 failures / 16 successes — roughly one success in six), that produced one
+/// lucky probe followed by **4712 exposed sessions in 30 days and no re-latch**:
+/// strictly worse than the permanent latch it replaced. The escalation ladder
+/// was only ever reachable by a model whose backend was completely gone, which
+/// is not the class either incident belongs to.
+///
+/// `1` — a single trip re-quarantines. A trip still costs
+/// [`CIRCUIT_BREAKER_THRESHOLD`] consecutive breaker-eligible failures, so a
+/// genuinely recovered model (which produces none) sails through probation with
+/// full throughput, while an intermittently-broken one is back in quarantine
+/// within a handful of sessions — and at the NEXT tier, because the tier is
+/// preserved across the release.
+const HARD_DISABLE_PROBATION_TRIP_CEILING: usize = 1;
+
+/// Consecutive successful sessions a bucket on probation must produce before the
+/// hard-disable history is forgiven and it returns to the ordinary
+/// [`TRIP_RATE_CEILING`] with a reset quarantine tier.
+///
+/// Deliberately evidence-based rather than clock-based: "prove you work" is the
+/// question probation asks, and a wall-clock window would forgive a model that
+/// was merely idle. `5` separates the two incident profiles cleanly — the
+/// recovered model from the 4-day outage went 10-for-10 immediately and clears
+/// probation on its fifth session, while a one-success-in-six model has a
+/// ~1-in-7776 chance of ever clearing it and therefore stays permanently one
+/// trip away from re-quarantine.
+///
+/// Runtime-only counter (`consecutive_successes`): a restart restarts the streak,
+/// which only ever keeps a bucket on probation LONGER. That is the safe
+/// direction.
+const HARD_DISABLE_PROBATION_SUCCESSES: u32 = 5;
+
 /// Quarantine length for an escalation `tier`: [`HARD_DISABLE_QUARANTINE_BASE`]
 /// doubled once per prior failed probe, capped at [`HARD_DISABLE_QUARANTINE_MAX`].
 fn quarantine_for_tier(tier: u32) -> Duration {
@@ -328,6 +370,21 @@ pub struct ModelHealth {
     /// pre-half-open snapshots, which load at tier 0 (the base quarantine).
     #[serde(default)]
     pub hard_disable_probe_tier: u32,
+    /// Seconds until the next half-open probe dispatch is admissible. `None`
+    /// when the bucket is not hard-disabled; `Some(0)` when a probe is
+    /// admissible right now. This is the field that answers "will this recover
+    /// on its own, and when?" — the question an operator could not answer for
+    /// four days. Derived, like `cooldown_seconds_remaining`, so `restore_all`
+    /// ignores it and re-anchors from the persisted tier instead.
+    #[serde(default)]
+    pub hard_disable_probe_seconds_remaining: Option<u64>,
+    /// The bucket's quarantine was released by a successful probe but it has not
+    /// yet produced enough clean sessions to have its hard-disable history
+    /// forgiven: it is dispatchable at full throughput, but a single breaker
+    /// trip re-quarantines it one rung higher. `#[serde(default)]` for
+    /// back-compat with pre-half-open snapshots.
+    #[serde(default)]
+    pub hard_disable_on_probation: bool,
     /// Number of breaker trips currently inside the rolling [`TRIP_RATE_WINDOW`].
     /// Surfaced so operators can see how close a bucket is to the hard-disable
     /// ceiling. `#[serde(default)]` for back-compat with pre-ceiling snapshots.
@@ -388,12 +445,47 @@ struct ModelState {
     hard_disable_probe_tier: u32,
     /// Whether a half-open probe has been handed out and has not yet been
     /// resolved. Set by `note_dispatch_accepted`, cleared by the first
-    /// failure/success that follows. Its only job is to make the tier escalate
-    /// **once per probe** rather than once per failure record — a single bad
-    /// session can produce both a stall and a failure. Runtime-only: a restart
-    /// drops it, which at worst costs one un-escalated quarantine, and
-    /// `restore_all` re-anchors the deadline anyway.
+    /// failure/success that follows.
+    ///
+    /// This is the **authority gate on quarantine transitions**, and nothing
+    /// else may move them:
+    ///
+    /// * Only an admitted probe's success can release the latch. A success from
+    ///   a session that was already in flight when the latch was set proves
+    ///   nothing about the quarantine — for a *flapping* model such successes
+    ///   are routine, and letting one release the quarantine having admitted
+    ///   zero probes hands the lane back for free.
+    /// * Only an admitted probe's failure can push the deadline out. Not every
+    ///   dispatch path consults the breaker (e.g. evidence-dispatch recovery
+    ///   reaches the pool without a `HealthTracker` read at all), so a
+    ///   health-blind path failing every few hours would otherwise re-anchor the
+    ///   quarantine forever and produce a silent permanent outage — the exact
+    ///   bug this whole change exists to remove, reintroduced by a different
+    ///   route. Measured: one failure record every 5h admitted **zero** probes
+    ///   in 30 simulated days, with the tier pinned at 0 so nothing showed it.
+    /// * It also makes the tier escalate once per *probe* rather than once per
+    ///   failure record, since a single bad session can emit both a stall and a
+    ///   typed failure.
+    ///
+    /// Note this gate does NOT weaken the fail-closed bound: exposure is bounded
+    /// by `note_dispatch_accepted` pushing the deadline at accept time, before
+    /// any verdict exists.
+    ///
+    /// Runtime-only: a restart drops it, which at worst costs one un-escalated
+    /// quarantine, and `restore_all` re-anchors the deadline anyway.
     probe_outstanding: bool,
+    /// The latch was released by a successful half-open probe, but the bucket
+    /// has not yet re-proven itself with [`HARD_DISABLE_PROBATION_SUCCESSES`]
+    /// clean sessions. While set, a single trip re-quarantines the bucket at the
+    /// NEXT tier (see [`HARD_DISABLE_PROBATION_TRIP_CEILING`]).
+    ///
+    /// Persisted, because forgetting it across a restart is what would let an
+    /// intermittently-broken model reset its history by waiting for a deploy.
+    hard_disable_on_probation: bool,
+    /// Consecutive successful sessions with no intervening failure. Drives the
+    /// probation exit only. Runtime-only; a restart restarts the streak, which
+    /// keeps a bucket on probation longer (the safe direction).
+    consecutive_successes: u32,
     /// Monotonic timestamps of recent breaker trips, pruned to
     /// [`TRIP_RATE_WINDOW`]. When its length reaches [`TRIP_RATE_CEILING`] the
     /// bucket is hard-disabled. Only genuine (escalating) trips are recorded —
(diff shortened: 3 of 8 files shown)


fernandobandeira:
## Review round 2 — all five defects fixed, each with a measured test

The review was right, and the framing was the most useful part: **the escalation ladder was only ever reachable by a model whose backend was completely gone.** The intermittently-broken model — the class *both* production incidents actually belong to — walked straight past it. My 30-day bound was real but vacuous for the input that matters.

Pushed as `aa9bc149e`.

### D1 — a lucky probe laundered the whole history

`clear_hard_disable` zeroed the tier, and `record_success` already cleared `trip_times` and `disable_ttl_trips`, so one successful probe left the bucket byte-identical to a never-tripped model.

**Fix: probation.** A probe success now calls `release_quarantine_to_probation` — the lane reopens at **full throughput** (that is what fixes the outage), but the escalation tier is preserved and while on probation a **single** trip re-quarantines one rung higher (`HARD_DISABLE_PROBATION_TRIP_CEILING = 1`) instead of eight. The history is forgiven only after `HARD_DISABLE_PROBATION_SUCCESSES = 5` consecutive clean sessions.

Evidence-based rather than clock-based, deliberately: "prove you work" is the question probation asks, and a wall-clock window would forgive a model that was merely idle. `5` separates the two incident profiles cleanly — the outage's model went 10-for-10 and clears probation on session five; a one-success-in-six model has a ~1-in-7776 chance of ever clearing it and so stays permanently one trip from re-quarantine. Both probation and the tier are persisted, so a deploy cannot launder them either.

A trip still costs `CIRCUIT_BREAKER_THRESHOLD` consecutive failures, so a genuinely recovered model produces none and is never penalised for being on probation.

### D2 — success without an admitted probe

`record_success` now returns early on a quarantined bucket unless `probe_outstanding`. Only an admitted probe can release a quarantine, and a stray success no longer wipes the trip window or the cooldown ladder either. You were right that "a dead model cannot manufacture a success" does not cover a *flapping* model.

### D3 — health-blind failures postponing the probe forever

`fail_probe` → `resolve_probe_failure`, which no-ops unless a probe is outstanding. This does **not** weaken the fail-closed bound you flagged as worth keeping: exposure is still bounded by `note_dispatch_accepted` pushing the deadline at accept time, *before* any verdict exists. The two concerns turned out to be independent.

### D4 — transient faults escalating

`record_transient_failure` resolves the probe with `escalate = false`: another quarantine period, not a longer one. The crate's own position (`TRANSIENT_BREAKER_THRESHOLD` = 20 vs `CIRCUIT_BREAKER_THRESHOLD` = 3) plus the fact that the original latch was minted *during* a transient incident makes this the right default at the one moment the model is being re-examined. A genuine failure on the next probe still escalates, asserted in the same test.

### D5 — the operator surface

Agreed the golden fanout was worth it. `ModelHealthOutput` now carries `hard_disable_probe_seconds_remaining` (`Some(0)` = probe admissible now, `None` = not quarantined), `hard_disable_on_probation` and `hard_disable_probe_tier`, and the `hard_disabled` doc no longer claims "no auto-expiry until a human re-enables". Regenerated via `make tool-goldens`; `make tool-goldens-check` and `pnpm mcp:types:check` both pass.

## The required test, measured

`an_intermittently_succeeding_model_cannot_launder_its_quarantine_history` — 30 simulated days, one-minute polling, one success in six:

| | admitted sessions in 30 days |
|---|---|
| `origin/main` | **0** (permanent latch, 4-day outage) |
| this PR before the probation fix | **42840**, never re-latches |
| this PR | **17**, ends quarantined at tier ≥ 1 |

Admission minutes: `[360,361,362,363, 1083, 2523, 5403,5404,5405,5406, 11166, 21246, 31326,31327,31328,31329, 41409]` — the design visible in the data: a probe that wins reopens the lane onto probation, three failures re-trip it, and it goes back in one rung higher (6h → 12h → 24h → 48h → 96h → 7d).

Also added: `a_success_without_an_admitted_probe_cannot_release_a_quarantine` (D2), `health_blind_failures_cannot_postpone_the_probe_forever` (D3 — asserts the noisy run admits *exactly* the same probes as a quiet baseline, not a threshold), `a_transient_upstream_fault_re_quarantines_without_escalating` (D4), `model_health_reports_the_next_probe_deadline_to_operators` (D5), `probation_survives_a_restart_so_it_cannot_be_laundered_through_a_deploy`.

### Surgical revert per fix

| reverted | tests that fail | measured |
|---|---|---|
| D1 probation release | 3 | 42840 sessions, no re-latch |
| D2 `probe_outstanding` gate on success | 1 | latch released with 0 probes admitted |
| D3 `probe_outstanding` gate on failure | 13 | 0 probes vs 6 for the quiet baseline |

Restored: **63/63 pass**.

## Test results

| Command | Result |
|---|---|
| `-p djinn-provider --lib catalog::health` | **63 passed / 0 failed** |
| `-p djinn-provider --lib` (full) | **625 / 0** |
| `-p djinn-control-plane` | **1019 / 0** across 10 targets |
| `-p djinn-server --lib server::tests::tool_schemas` | **18 / 0** |
| `-p djinn-provider --test tool_schema_projection_corpus --test tool_schema_corpus_loader` | **16 / 0** |
| `cargo clippy --workspace --all-targets --features qdrant --keep-going` | clean |
| `make tool-goldens-check`, `pnpm mcp:types:check` | pass |
| raw-SQL boundary + self-test | OK / 59 passed 0 failed |

Private PostgreSQL on port 55137, torn down.

### Correcting two of my earlier claims

**The 44 "pre-existing failures" in `djinn-provider --lib` were not pre-existing failures.** They were purely `TEST_POSTGRES_URL` not being set. With a database available the crate is **625/625**. My earlier framing was wrong and I should have checked that before reporting it as a known-bad baseline.

**The "2 identical failures, known `--test-threads` flake" claim was not established, and you were right to reject it.** Re-run under the same invocation this round, `-p djinn-coordinator --lib` produced **1** failure (`supervisor_impl::pr::tests::supported_active_explicit_legacy_adopts_through_every_poller`) rather than the previous 2, and a different test. It passes in isolation, and `origin/main` under the identical invocation on the same box produced the **same single failure**. So: the failing set varies run to run and with machine load, every member passes in isolation, and it is identical between this branch and main when measured back to back. That is the most I can honestly claim — it is evidence the suite is load-sensitive, not evidence of a specific known flake.

## Still not verified

* No staging/production run.
* The `evidence_dispatch_recovery` path that motivated D3 is now *harmless* to the quarantine, but it still dispatches without consulting the breaker at all. That is a pre-existing gap in a different module and out of scope here; it deserves its own fix.
* `HARD_DISABLE_PROBATION_SUCCESSES = 5` is reasoned, not empirically tuned against production data.

fernandobandeira:
## Review round 3 — N1–N6 fixed, exposure table now in the PR body

Pushed as `afb0f91b4`. The framing that mattered: **probation was defeated by failure clustering, not failure rate, and my test asserted against the one input where it worked.** A bound measured at a single point on the distribution proved nothing about any other point — which is precisely how the hole survived round 2.

### N1 — probation is now a counted, bounded trial

Re-quarantine fired on a *trip*, and a trip needs `CIRCUIT_BREAKER_THRESHOLD` **consecutive** breaker-eligible failures with any interleaved success resetting the counter, so `effective_trip_ceiling() == 1` was never even evaluated for a model that never failed three times in a row.

Probation is now: **win `HARD_DISABLE_PROBATION_SUCCESSES` (5) of at most `HARD_DISABLE_PROBATION_TRIAL_SESSIONS` (6)**. Successes and failures are counted independently and neither resets the other, so clustering is irrelevant. Three resolutions, all inside the budget: 5 wins → forgiven; `probation_failure_ceiling()` (derived: 2) losses → re-quarantined one rung higher; budget spent → re-quarantined.

On your point about the **unbounded window** (`adv_probation_window_worst_case_sessions` hitting 10,000): the third clause is charged in `note_dispatch_accepted`, i.e. on **exposure** rather than outcome, so it holds even when sessions never report anything. Exposure per released quarantine is `1 + 6` **by construction, against any input**. `a_probation_trial_is_bounded_by_dispatches_even_with_no_outcomes_at_all` asserts exactly this with zero outcomes recorded.

### N2 — the ledger no longer pre-charges

You were right that the D2 gate and the probation exit contradicted each other inside one function. The ledger is reset by `release_quarantine_to_probation` and advanced only by `charge_probation_*`, which no-op unless the bucket is actually on probation. The probe's own success is not charged to the trial either. Regression test added exactly as specified: 4 stray successes + 1 probe success → `probation=true`, tier preserved.

### N3 — absolute wall-clock deadline, per Fernando's call

`ModelHealth::hard_disable_probe_at_unix`; `restore_all` restores the remainder from it, so downtime counts as uptime would. Clamped to the tier's quarantine so a skewed snapshot can shorten but never manufacture one. `repeated_restarts_inside_the_quarantine_cannot_starve_the_probe` asserts the noisy run admits **exactly** the probes the quiet one does, at 1h/4h/5h cadences — the starvation direction my old test was blind to.

### N4 — probation tightens the throttle ladder

The trial counts every failure class including `record_stall(escalate=false)`, so the flap costs **2** sessions rather than 6. `probation_tightens_the_throttle_ladder_too` asserts the count equals `probation_failure_ceiling()` and explicitly bounds it below `PERSISTENT_THROTTLE_TRIP_THRESHOLD`.

### N5 — persistence is now a property production provides

Added `AppState::run_model_health_persist_loop`: 5-minute cadence plus a final snapshot on cancellation. The old test name was overclaiming, so it is renamed to `probation_survives_the_persist_restore_round_trip` — what it actually proves — with a comment pointing at the loop that makes that round-trip happen on a deploy.

### N6 — arithmetic corrected

Trial length changed the sum again. A one-success-in-six model must win ≥5 of 6: `C(6,5)·(1/6)^5·(5/6) + (1/6)^6 ≈ 0.00066`, **about 1 in 1500**. Verified numerically (1 in 1505).

## The exposure table (now in the PR body)

| pattern | this PR | round-2 (clustering hole) | `origin/main` |
|---|---|---|---|
| one success in 2 | **32** · quarantined · tier 5 | 42840 · never re-quarantined | 0 |
| one success in 3 | **24** · quarantined · tier 5 | 42840 · never re-quarantined | 0 |
| one success in 4 | **16** · quarantined · tier 5 | 32 | 0 |
| one success in 6 | **12** · quarantined · tier 5 | 17 | 0 |
| F,F,S repeating | **20** · quarantined · tier 5 | 40682 · never re-quarantined | 0 |
| S,S,F repeating | **48** · quarantined · tier 5 | 42840 · never re-quarantined | 0 |
| alternating S,F | **32** · quarantined · tier 5 | 42840 · never re-quarantined | 0 |
| all failures | **8** · quarantined · tier 5 | 8 | 0 |

Four properties asserted per row: bounded exposure, ends quarantined, escalated past tier 0, and still probed at all. I added `S,S,F` and `alternating S,F` beyond your list — `S,S,F` is the worst case at 48, and it is the one a rate-based intuition would have called "mostly fine".

Reverting N1 reproduces your measurements **exactly**: 42840 / 42840 / 32 / 17 / 40682. Reverting N3 gives `0 probes vs 8` under an hourly deploy. Reverting N2 gives the full pardon.

## Counts

`catalog::health` **68/0** · `djinn-provider --lib` **630/0** · `djinn-control-plane` **1022/0** · `djinn-server --lib` **665/0** · clippy workspace clean · `make tool-goldens-check` up to date · raw-SQL guard OK / self-test 59-0. Private PG on **55219**, torn down. All work stayed in the worktree this time.

`djinn-coordinator --lib` came back 2308/1 (`supported_active_explicit_legacy_adopts_through_every_poller`), passing in isolation and identical on `origin/main` under the same invocation on this box. Consistent with your read that these are load-sensitive rather than a specific known flake.

## Still not verified

* No staging/production run.
* The persist loop compiles and is wired, but its cadence and shutdown write have no integration test — `djinn-server` has no harness that boots `AppState` and advances time. Flagging rather than claiming it.
* `5` and `6` are reasoned, not tuned against production data. Per your note I fixed the counting and the clustering rather than tuning the constant; the binding constraint is now the trial length.
* `evidence_dispatch_recovery` still dispatches without consulting the breaker at all. It can no longer corrupt the quarantine, but the gap is pre-existing and in another module.

fernandobandeira:
## Parking this — three adversarial reviews, three rejections, each on a different input class

Recording the findings so the work isn't lost. **Not merged. Prod is unaffected and safe without it.**

| round | mechanism | defeated by | exposure over 30 simulated days |
|---|---|---|---|
| 1 | probe success → `clear_hard_disable()` | any success wipes tier + history | 4712, never re-latched |
| 2 | probation, tier preserved, re-quarantine on a *trip* | **failure clustering** (a trip needs 3 *consecutive* failures; any success resets) | 1-in-2 → **42840**, 1-in-3 → 42840, F,F,S → 40682 |
| 3 | counted trial "win 5 of at most 6", charged on exposure | **success clustering, manufactured by the quarantine itself** | token-bucket cap 6 → 6252; cap 16 → **42840**, never quarantined |

`origin/main` exposes **0** in every one of those cases.

### The round-3 finding is structural, not a distribution artifact

A quota-limited backend — the 429-throttle class this breaker exists for — *refills during the quarantine*. `HARD_DISABLE_QUARANTINE_BASE` is 6h, so at probe time it has ≥6 tokens banked: exactly `1 probe + HARD_DISABLE_PROBATION_SUCCESSES`. It wins every trial and earns a full pardon. The bucket-capacity cliff sits precisely at 6:

| bucket cap | 1 token/h | 4 tokens/h |
|---|---|---|
| 1–5 | 24 / 32 / 40 / 48 / 56 — all ✓ quarantined, tier 5 | same |
| **6** | **6252** ✗ tier 0 | **22445** ✗ tier 0 |
| 7–16 | 6251 ✗ | 22446 ✗ |

**Longer quarantines make it worse** — more banked quota. The underlying problem: *"did it succeed?" is a poor proxy for "is it healthy?" when the failure mode is quota exhaustion.* A rate-limited model always succeeds right after a wait, so any time-based quarantine feeds the thing it is trying to catch.

### Two changes here would regress prod relative to `main`

- **Persist loop is not leader-gated.** Spawned from `AppState::initialize()` on every replica. Helm is `replicas: 1` + `maxSurge: 1` → exactly 2 pods on every rolling deploy, both writing one global key, last-writer-wins. A surge pod's frozen boot snapshot can silently revert a human `model_health enable`. On `main` this cannot happen. Compare `graph_retention.rs` / `kueue_workload_reconcile.rs`, which carry enforcement tests requiring composition only in `become_leader`.
- **A re-quarantined model reports healthy to Prometheus.** `enter_quarantine` sets `hard_disabled` but not `auto_disabled`, and `breaker_metric_snapshot` keys off `auto_disabled`. After an incident whose post-mortem was "the alarm fired ~125×/day and nobody was notified", shipping a state where the alarm doesn't fire at all is a regression on that exact axis.

### Test-integrity notes worth keeping

- Deleting `task_dispatch.rs:1101` — the single line wiring `note_dispatch_accepted` — left the coordinator suite at **2307 pass / 2 fail, identical to unmodified HEAD**. The feature was unwirable with the suite green.
- Each round's flagship bound test asserted at effectively **one point on the distribution**, with a ceiling loose enough that neutering half the mechanism still passed. A bound measured at one input proves nothing anywhere else.
- The implemented rule was "win the first 5 consecutively", not "win 5 of at most 6" (`charge_probation_dispatch` fires before outcome 6 exists), so a model at 83% health was permanently quarantined — and the regenerated tool-schema goldens describe the *round-2* mechanism, which no longer exists.

### Why this is low urgency

The 4-day outage was caused by a **single-model `implement` lane**, not by the latch. With fallbacks now on all three lanes plus the VPS dispatch watchdog, a latched model is a degradation rather than an outage.

If revisited, the simpler shape is a plain long TTL on the hard-disable (e.g. 24h, no probation, no trial): a bounded re-probe cost for a genuinely dead model, and no clustering surface at all.

fernandobandeira:
Closing. Three adversarial reviews, three rejections, each defeated by a different input class — the detailed findings are in the comment above and stay on the record here.

The deciding point: the 4-day outage was caused by the single-model `implement` lane, not by the latch itself. That cause is fixed and deployed, and with fallbacks on all three lanes plus the VPS dispatch watchdog, a latched model is now a degradation rather than an outage. This change would have traded that for a measured regression (up to 42,840 sessions exposed where `main` exposes 0, a non-leader-gated persist loop able to revert a human `model_health enable`, and a re-quarantined model reporting healthy to Prometheus).

Anyone picking this up later should start from the structural finding rather than the mechanism: "did it succeed?" is a poor proxy for "is it healthy?" when the failure mode is quota exhaustion, because a rate-limited model always succeeds right after a wait. Any success-counting scheme behind a time-based quarantine — including a plain long TTL — is defeated by the same refill. It likely needs a different signal entirely (provider rate-limit headers, error class, latency), not another variation on counting.

The adversarial sweep harness that produced every number reproduces by dropping `adv_review_sweep.rs` into `server/crates/djinn-provider/tests/`.

PR #3182, opened 2026-08-17
fix(pr-poller): observe a merged PR before every advance gate (4vnt/#3153, 3kza/#3155)

fernandobandeira:
A PR merges, the commit is on `main`, and the djinn task stays at `pr_draft` forever with `closed_at: null` and `merge_commit_sha: null`. It never self-heals. The damage is not the status column — **a task that never closes never releases its blockers**, so a whole dependency chain silently drops out of the build loop.

Observed twice in production on 2026-08-16.

| task | PR | merged | stranded | dependent |
|---|---|---|---|---|
| `4vnt` | #3153 | 2026-08-11T13:22:12Z | **5 days** in `pr_draft`, all 5 ACs met, `ci_status: passing` | `296y`, blocked the whole time; dispatched within seconds of a manual close |
| `3kza` | #3155 | 2026-08-16T23:14:10Z | `pr_draft` with `ci.last_seen_at: 23:14:54` — the poller ran **44 s after** the merge | `i58q`, released the instant the task was closed by hand |

## Primary root cause

`poll_pr_draft_tasks` evaluated the **tripwire active-hold gate before reading `pr.merged`**.

`server/crates/djinn-coordinator/src/pr_poller/pr_watcher.rs` (pre-fix line numbers):

```
116  let ci_status = self.record_ci_snapshot(...).await;   // advances ci.last_seen_at
132  if self.reconcile_tripwire_hold(&task, pull_number, &pr.head.sha).await {
141      continue;                                          // ← masks everything below
     }
144  if pr.merged == Some(true) { ... }                     // never reached
172  if pr.state == PrState::Closed { ... }                 // never reached
```

Both tasks carry a `tripwire.gate.held` activity row on the PR's **live** head, with `severity: human_review_required` and no matching release — so `reconcile_tripwire_hold` returned `true` on every tick, forever:

* `4vnt` — held on `d41958f6…` at 2026-08-09T16:56:24Z (3 × `dependency_identity_change`); PR merged 2026-08-11.
* `3kza` — held on `fef4a526…` at 2026-08-16T22:55:55Z (3 × `large_delete_or_rewrite`); PR merged 19 min later.

`ci.last_seen_at` advancing *past* the merge is what pins the location: the loop reached `record_ci_snapshot` (line 116) and then `continue`d, and the only statement between that call and the merged branch is the tripwire hold. `poll_pr_review_tasks` has always had the correct order (merged → closed → CI gate → tripwire gate); only the `pr_draft` loop was inverted.

### Hypotheses ruled out

| hypothesis | verdict | evidence |
|---|---|---|
| merge detection keyed off a stale stored SHA | **no** | the branch is `pr.merged == Some(true)` from the live fetch; no SHA is compared. `ci_head_sha` matched the merged head exactly on both tasks (`d41958f6`, `fef4a526`). |
| the merge queue merges a different commit than the PR head | **no** | same — no SHA comparison exists on this path. |
| `task_pr_handling_is_eligible` / direct-delivery gating short-circuits | **no** | that check (line 41) precedes `record_ci_snapshot` (line 116), and `ci.last_seen_at` kept advancing — so eligibility passed on every tick. |
| the `pr_draft` loop has no merge-terminalization branch at all | **no** | it exists at line 144; it was simply unreachable under a hold. |

The `github_head_sha` divergence noted during triage (`b3aaede1…` vs `d41958f6…`; `0d67ee6e…` vs `fef4a526…`) is real — both branches were force-pushed during a rebase — but it is **not** this bug's cause. It is preserved as a regression anyway, because a future SHA-keyed merge check would strand these same tasks for a second, independent reason.

`merge_commit_sha: null` on both tasks is a consequence of the operator's manual close (a plain `pr_draft → closed` status change), not a separate defect: `apply_pr_merge` already persists it. The reconciliation path now asserts it.

## The safety-net hole

`merged_reconcile.rs` exists precisely for "a PR merged and nobody noticed". Its query, `TaskRepository::list_reconcilable_pr_tasks` (`djinn-db/src/repositories/task/reads.rs`), read:

```sql
WHERE pr_url IS NOT NULL AND pr_url <> ''
  AND status NOT IN ('closed', 'pr_draft', 'pr_review')
```

It excluded `pr_draft` and `pr_review` — the exact statuses both incidents were stuck in — on the theory that the poller loops kept sole ownership there. That put the hole in the net precisely where the primary mechanism operates, so a miss in the poller loop was covered by nothing and never self-healed.

## What changed

**Primary** (`pr_watcher.rs`) — **only** the merged branch moves above the tripwire hold, driven by a new shared classifier `merged_reconcile::classify_pr_terminal_state(merged, state)`. It takes no SHA and no gate state, so neither a hold nor a stored head can gate merge detection. The reconciler's `decide_blindspot_reconcile_action` delegates to the same classifier, so the primary path and the safety net cannot drift apart on what "merged" means.

The closed-unmerged branch deliberately **stays below** the hold, byte-identical to `main`. Merging is a fact Djinn records — once GitHub says merged, every advance action below the gate is moot. Force-closing is an action Djinn *takes*, and `ForceClose` is exempt from the blocks-others guard (`task/status.rs:634`), so it releases every dependent. Hoisting it would mean a reviewer who closes a human-review-held PR unmerged, intending a redo, instead releases dependents onto work that never landed — a wrong-close, worse than the stranding this fixes, and not required by either incident. The ordering guard now pins the asymmetry in both directions.

**Safety net** (`reads.rs`, `merged_reconcile.rs`) — the query becomes `status <> 'closed'`. Terminalization routing splits: a task already in `pr_draft`/`pr_review` takes `PrMerge` directly (carrying `RECONCILE_MERGE_REASON` on the `pr_merge` transition via the new `apply_pr_merge_with_reason`), skipping the pointless `user_override → pr_review` hop; blind-spot statuses keep that hop, since `PrMerge` is only valid from those two. Idempotency: the task row is re-read immediately before the transition and a `closed` task is skipped, so the poller and the reconciler observing the same merge cannot double-transition. Open PRs in poller-owned statuses are left completely alone — their own loop refreshes CI every tick.

## Tests

Each of the first six was verified to **fail against the pre-fix code** and pass after (stash-and-verify runs in the task report).

| test | proves |
|---|---|
| `merged_pr_under_active_tripwire_hold_still_closes_its_pr_draft_task` (`supervisor_impl/pr_tests.rs`) | **End-to-end.** Drives the real `poll_pr_draft_tasks` — real installation-token client, real CI snapshot recording, real tripwire hold reconciliation, only the HTTP endpoint redirected via the existing wiremock seam — against a PR GitHub reports as `merged: true` with an active hold on the live head *and* a stale stored head SHA. **Pre-fix it observes `pr_draft`**; post-fix `closed` / `completed` / populated `merge_commit_sha`, and the blocked dependent becomes ready. |
| `a_merged_pr_is_observed_before_the_tripwire_hold_gate` | Source-level ordering guard, honestly labelled, with named failing mutations — the loop cannot regress into gating merge detection again. Verified to fail when the tripwire block is moved back above the terminal match. |
| `list_reconcilable_covers_poller_owned_statuses` | The safety-net hole: `pr_draft`/`pr_review` tasks with a PR are reconcilable. |
| `merged_pr_in_pr_draft_terminalizes_and_releases_its_dependent` | The 4vnt/3kza shape through the reconciler: select → decide → terminalize; `close_reason: completed`, populated `merge_commit_sha`, reconciliation named in the audit trail, **and the blocked dependent becomes ready**. |
| `force_pushed_stale_head_shas_do_not_block_terminalization` | Stored snapshot head and `task_attempts.github_head_sha` both stale against the merged head (the production divergence, verbatim SHAs) — the task still terminalizes. |
| `merged_pr_in_a_blind_spot_status_routes_through_pr_review_and_closes` | The `user_override → pr_review` hop for a non-poller-owned status (`needs_task_review`) — the status `poll_pr_review_stuck_tasks` leaves a merged PR sitting in, since that loop observes the merge and merely `continue`s. Verified by deleting the hop: 12 passed, 1 failed, only this test. |
| `reconciled_merge_is_idempotent` / `poller_terminalized_task_is_left_alone` / `list_reconcilable_excludes_terminalized_tasks` | Running twice, and running against a task the poller already terminalized, causes no second transition and no extra audit row. |
| `closed_unmerged_logs_only`, `open_pr_refreshes_ci_only` (preserved) | Negative controls: a CLOSED-unmerged PR still does not close its task; an OPEN PR is still left alone with only a CI refresh. |

**Results:** `cargo nextest run -p djinn-coordinator --lib --features test-support` → **2319 passed, 0 failed**. `cargo fmt --all`, `git diff --check`, `cargo clippy` (djinn-db + djinn-coordinator, lib + tests) all clean. `BASE_SHA=origin/main ./scripts/check-raw-sql-boundary.sh` → OK (6 files checked); `./scripts/test-check-raw-sql-boundary.sh` → 59 passed, 0 failed. All DB-backed tests ran against a dedicated throwaway Postgres, never the shared test database.

### DIFF
--- a/server/crates/djinn-coordinator/src/pr_poller/ci_routing/executor/tests.rs
+++ b/server/crates/djinn-coordinator/src/pr_poller/ci_routing/executor/tests.rs
@@ -6766,8 +6766,12 @@ fn the_pr_poller_closes_ci_routes_on_both_merge_and_pass() {
     );
 
     // ── Each call site sits in the branch that needs it ─────────────────────
+    //
+    // The merged branch is the `Merged` arm of the terminal-state match that
+    // now runs BEFORE the tripwire active-hold gate (the `4vnt`/`3kza` fix); a
+    // merged PR is ground truth and no Djinn-side gate may precede it.
     let merged_branch = code
-        .find("if pr.merged == Some(true) {")
+        .find("merged_reconcile::PrTerminalState::Merged => {")
         .expect("the merged branch is in pr_watcher.rs");
     let apply_merge = code
         .find("self.apply_pr_merge(")
--- a/server/crates/djinn-coordinator/src/pr_poller/ci_routing/executor/tests.rs
+++ b/server/crates/djinn-coordinator/src/pr_poller/ci_routing/executor/tests.rs
@@ -6725,8 +6725,7 @@ async fn the_quiescence_report_counts_the_leaders_live_provider_futures() {
 ///     `["true", "false"]` assertion fails — and a merged PR would record
 ///     `passed`, or a passing one `merged`.
 /// (e) Move the merged-branch call after `apply_pr_merge`, or out of the
-///     `pr.merged == Some(true)` branch entirely: the first range assertion
-///     fails.
+///     `PrTerminalState::Merged` arm entirely: the first range assertion fails.
 /// (f) Move the passing-branch call out of the `CiStatus::Passing` arm: the
 ///     second range assertion fails.
 /// (g) Shadow the lane-routing method with a local `fn` of the same name in
--- a/server/crates/djinn-coordinator/src/pr_poller/merged_reconcile.rs
+++ b/server/crates/djinn-coordinator/src/pr_poller/merged_reconcile.rs
@@ -1,38 +1,102 @@
-//! Blind-spot merged-PR reconciliation.
+//! Merged-PR reconciliation (the poller's safety net).
 //!
-//! The PR poller's status-scoped loops only poll `pr_draft` and `pr_review`
-//! tasks. A PR can merge — or its head CI can change — while its task sits in a
-//! status the poller never observes (`open`, `in_progress`, `needs_task_review`,
-//! `needs_lead_intervention`, ...). When that happens the merge is never noticed
-//! and, worse, the task's promoted CI snapshot can stay frozen at `failing`,
-//! which makes the respawn guard bypass open-PR adoption and dispatch rework
-//! workers against an already-merged PR.
+//! A PR can merge — or its head CI can change — without the status-scoped
+//! poller loops terminalizing the task. Two shapes cause that:
 //!
-//! This pass closes that blind spot. It runs on the slow stale-sweep cadence
-//! (never per tick) over the bounded set of non-terminal tasks that carry a
-//! `pr_url` and are NOT in a poller-owned status. For each it fetches PR state
-//! via the same installation-token GitHub plumbing the poller already uses:
+//! 1. The task sits in a status the poller never observes (`open`,
+//!    `in_progress`, `needs_task_review`, `needs_lead_intervention`, ...). The
+//!    merge is never noticed and, worse, the task's promoted CI snapshot can
+//!    stay frozen at `failing`, which makes the respawn guard bypass open-PR
+//!    adoption and dispatch rework workers against an already-merged PR.
+//! 2. The task IS in a poller-owned status (`pr_draft` / `pr_review`) but the
+//!    owning loop `continue`d past its merge check. This is not hypothetical:
+//!    incidents `4vnt`/#3153 and `3kza`/#3155 both stranded in `pr_draft`
+//!    because an active tripwire hold short-circuited the `pr_draft` loop
+//!    *before* its merged branch (see `pr_watcher::poll_pr_draft_tasks`). The
+//!    primary defect is fixed there, but this pass must cover those statuses
+//!    too: a net with a hole exactly where the primary mechanism operates is
+//!    not a net. A task that never terminalizes never releases its dependents,
+//!    so the blast radius is a silently-stalled dependency chain, not a wrong
+//!    status column.
+//!
+//! The pass runs on the slow stale-sweep cadence (never per tick) over the
+//! bounded set of non-terminal tasks that carry a `pr_url`. For each it fetches
+//! PR state via the same installation-token GitHub plumbing the poller uses:
 //!
 //! - MERGED → terminalize with the same merge bookkeeping the poller applies for
-//!   a `pr_review` merge (the `pr_merge` path), routed through
-//!   `user_override → pr_review` first because `PrMerge` is only a valid
-//!   transition from `pr_draft`/`pr_review`. This mirrors the manual
-//!   `user_override → pr_merge` an operator runs today.
+//!   a `pr_review` merge (the `pr_merge` path). From `pr_draft`/`pr_review` the
+//!   `PrMerge` transition applies directly; from any other status the task is
+//!   first routed through `user_override → pr_review` because `PrMerge` is only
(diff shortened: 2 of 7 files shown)


fernandobandeira:
Review addressed in `ef6a5c1a0`. Both Medium findings accepted; no pushback.

**1. `ClosedUnmerged` reverted below the tripwire gate.** Accepted as stated — hoisting it was unrequired scope in the wrong direction. `ForceClose` is in the `work_landed` exemption at `task/status.rs:634`, so it bypasses the blocks-others guard and releases every dependent; a reviewer closing a held PR unmerged to force a redo would have released dependents onto work that never landed. Merging is a fact Djinn *records*; force-closing is an action Djinn *takes*. The block is now byte-identical to `main` and in its original position — `git diff origin/main -- pr_watcher.rs` is one behavioural change, the merged block moving above the hold. Ordering (merged 166 → hold 199 → closed 214) is pinned in both directions by the source guard, with `hold_gate < force_close` as named mutation (e); verified failing when hoisted.

**2. Blind-spot hop now pinned.** `merged_pr_in_a_blind_spot_status_routes_through_pr_review_and_closes` drives the production `terminalize_reconciled_merge` on a `needs_task_review` task and asserts `closed`, `close_reason: completed`, `merge_commit_sha`, the `RECONCILE_MERGE_REASON` audit payload, and a released dependent. Verified your way — deleting the hop gives **12 passed, 1 failed**, only the new test, `left: "needs_task_review", right: "closed"`. That status is the concrete one: `poll_pr_review_stuck_tasks` (`pr_watcher.rs:921`) observes `pr.merged == Some(true)` and merely `continue`s, so this pass is the only thing that terminalizes such a task — the claim now rests on a test.

**Comment accuracy, all three:** the false "no HTTP double / no base-URL seam" claim is gone from both files and now points at `set_installation_client_base_url_for_test` naming the wiremock test as the preferred place to extend; `ci_routing` mutation (e) re-anchored off the deleted `pr.merged == Some(true)` literal; the fan-out wording corrected to say only the `record_ci_snapshot` write and its own follow-up calls are skipped (the PR fetch has already happened). I left the pre-existing identical prose in the `ci_routing` guard's own doc alone as out of scope.

Untouched as instructed: the orphaned `tripwire.gate.held` row, the `merge_commit_sha` reasoning, the re-anchored guard, and the DB-layer `FOR UPDATE` serialization.

**Verification:** `cargo nextest run -p djinn-coordinator --lib --features test-support` → **2319 passed, 0 failed** (2318 + the new test) on the committed tree. `cargo fmt --all`, `git diff --check`, `cargo clippy` (djinn-db + djinn-coordinator, lib + tests) clean. Post-commit: `BASE_SHA=origin/main ./scripts/check-raw-sql-boundary.sh` → OK, 6 files; `./scripts/test-check-raw-sql-boundary.sh` → 59 passed, 0 failed. Dedicated throwaway Postgres on port 55163, torn down.

